PR #557, opened 2013-06-05
Add hello world command to demo bundle and configure new log to console handler

Tobion:
First commit: allows the master branch to install 2.4-dev.
Second commit: added a hello world command in demo bundle
Third commit: configure the log to console handler that is possible since symfony/MonologBundle#42

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/symfony": "~2.4",
-        "doctrine/orm": ">=2.2.3,<2.4-dev",
+        "doctrine/orm": "~2.2,>=2.2.3",
         "doctrine/doctrine-bundle": "~1.2",
         "twig/extensions": "~1.0",
         "symfony/assetic-bundle": "~2.3",


app/config/config_dev.yml
```diff
@@ -14,15 +14,18 @@ web_profiler:
...
         main:
-            type:  stream
-            path:  %kernel.logs_dir%/%kernel.environment%.log
-            level: debug
+            type:   stream
+            path:   %kernel.logs_dir%/%kernel.environment%.log
+            level:  debug
+        console:
+            type:   console
+            bubble: false
```
lyrixx: Hello. 

I just hit an issue here with bubble and I would like to discuss about it

I was trying to debug my application, and so I added a new handler to push all my logs to the Elasticsearch/Heka/Kibana stack (to get a better visualisation / filter system for my logs...) . I did not really think about the order of the handlers and so I added my new handler after the previous handlers. And so I ended-up with the following config:

```
monolog:
    handlers:
        main:
            type:   stream
            path:   "%kernel.logs_dir%/%kernel.environment%.log"
            level:  debug
            channels: ["!event", "!consumer_status", "!beberlei_metrics"]
        console:
            type:   console
            bubble: true
            channels: ["!doctrine", "!event", "!consumer_status", "!beberlei_metrics"]
        heka:
            type: socket
            level:  debug
            connection_string: "udp://heka.lxc:5565"
            timeout: 1
            connection_timeout: 1
            formatter: monolog.formatter.json
```

After that, I tried to debug my code and noticed I missed lot of logs. I try to understand why (is it an issue with ES or heka or kibana ? ; It is an issue with my code ? ; etc ?). And I finally understood my issue.

So finally, I really think the console handler should bubble all logs, because It's "just" a nice to have feature (A very good one BTW, thanks again). But it should not "eat" logs. All handlers registered after the handler should also be able to get all logs. So, IMHO, we should turn on the bubble here.

WDTY ?

ping @Tobion @Seldaek

Pierstoval: It's all about Monolog configuration, so IMO the dev should read the docs and know that the order is important when defining log handlers as said in the [core concepts](https://github.com/Seldaek/monolog/blob/master/doc/01-usage.md#core-concepts) (dead link abofe)

Tobion: it doesnt really matter if bubble is enabled or not as the firephp and chromephp handler are not enabled anymore by default in SE.

lyrixx: @Tobion Yes I know and my "issue" was not about firephp. It was about DX ; to remove a possible WTF effect.

Tobion: feel free to change it

lyrixx: I already changed it in my app ; But I wanted to know if you think it's a good idea to change it in symfony too.

Tobion: I'm ok to change it

PR #658, opened 2014-05-09
Run $kernel->terminate after connection is closed in HHVM

lmcd:
In php-fpm, we can run cleanup code/blocking tasks after the HTTP connection is closed with `fastcgi_finish_request`. HHVM doesn't have this function, but it does have the counterpart `register_postsend_function`.

See here:
https://github.com/facebook/hhvm/blob/master/hphp/doc/extension.new_functions
https://github.com/facebook/hhvm/issues/1230

Note: haven't tested yet. Just putting it out there.

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -27,13 +27,10 @@ $request = Request::createFromGlobals();
 $response = $kernel->handle($request);
 $response->send();
 
-$postSend = function() use ($kernel, $request, $response) {
-    $kernel->terminate($request, $response);
-};
-
 if (function_exists('register_postsend_function')) {
-    register_postsend_function($postSend);
-}
-else {
-    $postSend();
+    register_postsend_function(function() use ($kernel, $request, $response) {
+        $kernel->terminate($request, $response);
+    });
+} else {
+    $kernel->terminate($request, $response);
 }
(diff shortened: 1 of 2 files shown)


allan-simon:
Hello, how can one test this ?

alesblaznik:
@lmcd I've tested it just now on HipHop VM 3.6.2 (rel) and it works as expected - not any issues so far.

@fabpot Should this be added to Symfony standard since a lot of folks are starting use HHVM nowadays and actually it took me a while to recognize that fast_cgi_request_finish don't work on hhvm? Maybe for now it could be enought that this lines should be commented and say "uncomment this, if you're using hhvm"

fabpot:
If the code works, let's see where this should be added in Symfony itself.

hacfi:
We could close this!? If it’s really missing someone will notice, test it and this can be merged.

fabpot:
As HHVM is not a priority for the core team, nobody is using it (we even discussed removing its official support in Symfony), someone using HHVM should finish this change so that we can merge it. In any case, it cannot be merge as is as this should be part of Symfony, not the standard edition.

fabpot:
Closing this one as apparently nobody has the issue... or nobody uses HHVM and Symfony.

PR #673, opened 2014-06-20
Use PSR-4

Tobion:
symfony/symfony#11189 for Symfony Standard

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "type": "project",
     "description": "The \"Symfony Standard Edition\" distribution",
     "autoload": {
-        "psr-0": { "": "src/", "SymfonyStandard": "app/" }
+        "psr-4": { "": "src/", "SymfonyStandard\\": "app/" }
     },
     "require": {
         "php": ">=5.3.3",


hacfi:
:+1:

WouterJ:
-1 for including `src/` in the psr-4 autoloading. We should leave that in psr-0 imo

Tobion:
@WouterJ please explain negative votes.

weaverryan:
@Tobion Is there any functionality difference at all with this? It seems like I'd still be able to put *anything* into `src/` and have it auto-loaded just like now.

I guess I'm asking, what's the motivation for this? But at the same time, what's the motivation *against* this if it keeps the same functionality as now?

WouterJ:
Hmm, seems like I understood PSR-4 wrongly. As it turns out, empty namespaces will work just like empty namespaces in PSR-0.

However, still having `src/Acme/Bundle/BlogBundle` instead of `src/BlogBundle` seems more "PSR-0" than "PSR-4". I'm wondering if we aren't learning people wrong things if we tell them `src/Acme/Bundle/BlogBundle` is PSR-4.

However, the remaining doubt won't stop me from changing my -1 into a neutral vote.

Pierstoval:
IMO  , the "pros" are that PSR-4 is loaded faster than PSR-0, the "cons" are that PEAR-like structures will be no longer supported with PSR-0.

> I'm wondering if we aren't learning people wrong things if we tell them src/Acme/Bundle/BlogBundle is PSR-4.
If PSR-4  is target as `"": src/`, then `src/Acme/Bundle/BlogBundle.php` will still be `Acme\Bundle\BlogBundle` and it won't change.

Tobion:
Btw, PSR-0 is deprecated by now anyway.

> Deprecated - As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative. - http://www.php-fig.org/psr/psr-0/

Tobion:
Merged #796 into 2.7 as this is also the branch where we adjusted symfony itself to use PSR-4. So this is consistent.

Tobion:
The lock file needs to be updated so the warning about out-of-sync does't appear. This will be done with the next beta release anyway.

PR #687, opened 2014-07-29
make config_dev.yml a valid YML file

jdespatis:
`%..%` must be wrapped by quotes, or YML file is not valid!

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -8,8 +8,8 @@ framework:
     profiler: { only_exceptions: false }
 
 web_profiler:
-    toolbar: %debug_toolbar%
-    intercept_redirects: %debug_redirects%
+    toolbar: "%debug_toolbar%"
+    intercept_redirects: "%debug_redirects%"
 
 monolog:
     handlers:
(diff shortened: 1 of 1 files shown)


jdespatis:
@stof according to the yaml 1.2 specs: http://www.yaml.org/spec/1.2/spec.html, the `%` is forbidden, it's a directive start symbol
(look at the `5.3. Indicator Characters` section)

@fabpot Is there some unit test somewhere that checks the validity of each yaml file of symfony-standard ?
I think it would be interesting to have this to be sure all yaml config files are valid

My yaml reader crashes for another yaml file for example: https://github.com/symfony/symfony-standard/blob/v2.6.4/app/config/routing.yml
due to the `@` symbol, which is forbidden, (reserved for futur use in yaml 1.2)

stof:
@jdespatis See http://www.yaml.org/spec/1.2/spec.html#id2788859 The plain style for scalars allows ``%``, because it is not ambiguous with directives

jdespatis:
@stof I've read this section, but I'm not sure to see where it's said that this `%` is allowed (I can however read the starting character `:  ? ~` are allowed on some circumstances)

Please could you quote to me the text that indicates it's possible ?

PR #702, opened 2014-08-23
[DX] Apply cssrewrite to css, sass and less files by default

peterrehm:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #670 |
| License | MIT |
| Doc PR | - |

Tweked the configuration to apply cssrewrite to css, sass and less files by default

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -36,7 +36,8 @@ assetic:
     bundles:        [ ]
     #java: /usr/bin/java
     filters:
-        cssrewrite: ~
+        cssrewrite:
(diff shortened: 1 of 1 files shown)


broncha:
Personally, I think its useless to have my scss files inside public directory. I like keeping them outside the public directory, that means i will have to use the @ notation. But cssrewrite does not work well with @ notation. The workaround for me is to setup a virtual host and give absolute url to images.

PR #736, opened 2014-11-14
fix .htaccess for apache to v2.2 or 2.3 and later

skilla:
mod_authz_core is authorization core, but the responsible of allow/deny access to host/ip/... is mod_authz_host or for compatibility reasons mod_access_compat, this is a reasonable aproach for a variety of installations.

### DIFF
--- a/app/.htaccess
+++ b/app/.htaccess
@@ -1,5 +1,13 @@
 <IfModule mod_authz_core.c>
-    Require all denied
+    <IfModule mod_authz_host.c>
+        Require all denied
+    </IfModule>
+    <IfModule !mod_authz_host.c>
+        <IfModule mod_access_compat.c>
+            Order deny,allow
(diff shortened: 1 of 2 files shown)


jrobeson:
ping @skilla ? did you test these ruiles on apache 2.2 and 2.4?

weaverryan:
And does the "old" approach not work (or being deprecated)? Since these `.htaccess` files are just here for security, I'd favor having something that always works and is clear. So, unless the existing code doesn't work on some new Apache version, I'd be :-1: on this

Tobion:
The require directive comes from mod_authz_core. 

> mod_authz_host extends the authorization types with ip, host and local. Other authorization types may also be used but may require that additional authorization modules be loaded.

But since we only use "Require all denied", we dont need mod_authz_host.

PR #737, opened 2014-11-27
Do strict check on environment name in AppKernel.php

gerryvdm:
Enable strict in_array() comparison when conditionally enabling development bundles.


Tobion:
This change doesn't add anything useful but also doesn't hurt. Maybe strict checking is even faster.
So lets just merge it. But should be done in 2.3.

@Pierstoval I think you are missing the point. It's only about `array('dev', 'test')`, so intergets will not match anyway.

fabpot:
Thank you @gerryvdm.

PR #739, opened 2014-11-28
Remove unused twig-extensions dependency

tommygnr:
The ```twig/extensions``` library is not actually used or enabled in the symfony standard distribution so its inclusion in the composer.json file seems unnecessary. I did quite a bit of digging around to work out why it was ever included and as best I can tell it was because in the betas of ```2.0``` there was a convenient way to register these extensions in the twig configuration block. That approach was abandoned in https://github.com/symfony/symfony/commit/96045739b198b030ab159425ce19614fe2e0e665


dunglas:
:+1:

fabpot:
Thank you @tommygnr.

PR #745, opened 2014-12-02
Updated checks for localhost to extend to RFC1918 too

udf2457:
Currently, you only check for localhost usage.

It may well be people are running in a console-only VM or on a server, in which case, allowing requests from RFC1918 or reserved subnets would be desirable.

The goal of this patch is to allow this to happen for app_dev.php and config.php.


Pierstoval:
I think this PR can be closed as it's no longer relevant for current Symfony versions

stof:
We already rejected doing such change in symfony-standard directly in the past. So -1

PR #757, opened 2014-12-12
Use getRootDir() rather than assuming __DIR__

thewilkybarkid:
In the event that `getRootDir()` in `AppKernel` is overridden (eg the file is moved elsewhere but the config etc stays put), the `registerContainerConfiguration` would also have to be changed.


jrobeson:
what's the holdup on this one? any reason not to have it?

fabpot:
Thank you @thewilkybarkid.

PR #760, opened 2014-12-16
Adding security advisories pre-update check via `roave/security-advisories` meta-package

Ocramius:
This additional package behaves like https://github.com/FriendsOfPHP/security-advisories, but before actually running the installation during any `require` or `update` composer command executions.

This effectively prevents installation of any security-issue affected code when changing dependencies.

The package itself does not bring in any dependencies or any additional code.

_must_ be referencing `dev-master`.

More on it:
- https://github.com/Roave/SecurityAdvisories
- [Roave/SecurityAdvisories release blogpost](http://ocramius.github.io/blog/roave-security-advisories-protect-against-composer-packages-with-security-issues/)

I didn't update the `composer.lock` since I'd first need some feedback on whether this change is actually welcome.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "~3.0,>=3.0.12",
         "sensio/framework-extra-bundle": "~3.0,>=3.0.2",
-        "incenteev/composer-parameter-handler": "~2.0"
+        "incenteev/composer-parameter-handler": "~2.0",
+        "roave/security-advisories": "dev-master"
     },
     "require-dev": {
         "sensio/generator-bundle": "~2.3"


willdurand:
:+1:

weaverryan:
My only worry would be the error message from Composer. If it give you the standard "cannot find a suitable package" type of message, that might be quite confusing. But if it gives you something clear about there being a security issue, that's different. I haven't tried this package yet, and I like it, but this was my concern when I read the blog post.

Also, a package+release I'm using suddenly has a security issue, I won't know about it until I try to update. Additionally, the error might occur when I'm doing something totally unrelated - like adding a new library, which could again be confusing.

People (me too) are lazy about security stuff, and I'd like to help "motivate" them :). This may or may not be the best way, based on my questions above (but maybe it is!).

Thanks!

Ocramius:
@weaverryan I totally get your concerns, but I don't have a different way to perform this. A different approach would be a post-install hook, but then it's already too late, no?

cordoval:
I think this makes a lot of sense :+1: 
Since the error on composer exposes the conflict between roave package of advisories and the version of symfony that in this case is vulnerable then I think it is understandable and there shouldn't be any confusion. I doubt people would try to make any other conclusion if the conflict is specifically to the roave package.

```bash
~ composer install                                        Luiss-iMac [20:24:01]
Loading composer repositories with package information
Installing dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for symfony/symfony 2.5.2 -> satisfiable by symfony/symfony[v2.5.2].
    - roave/security-advisories dev-master conflicts with symfony/symfony[v2.5.2].
    - Installation request for roave/security-advisories dev-master -> satisfiable by roave/security-advisories[dev-master].
```

Thanks for sharing @Ocramius! :older_man:

Pierstoval:
A little bit of documentation might be helpful enough to explain how this package works.
 
Or, if this package is not added to the standard edition, someone using it will *already* be aware of how it works, because he installed it himself.

Ocramius:
> A little bit of documentation might be helpful enough to explain how this package works.

As in added to symfony-docs?

Pierstoval:
If you already made a PR on the symfony-docs, could you link it in your first post for the two PR's consistency/comprehension ?

cordoval:
honestly this package is a no brainer. The documentation is on the blog post pretty much and the package's readme.md. We can certainly add docs to symfony documentation probably after it has been accepted/voted by the gatekeepers.

Ocramius:
Note: rebased.

rjmunro:
@Ocramius If you have a working site, waiting for composer update is already too late, changing to a post-install composer hook is not worse.

This will blocking an update that might fix some security vulnerabilities you have if it doesn't fix all of the vulnerabilities you have. Also the output is totally confusing, unless you know what is going on. I think this package is more likely to stop you upgrading than encourage you to upgrade, unless you are expert enough to know what it is.

I've added a PR for the post install hook we use where I work at #954.

Pierstoval:
Does this PR still needs review?

As we already have `bin/console security:check`, that could be automated in CI and deployment, is this feature really needed in Symfony Standard?

Moreover, as of SF4 + Flex, we might have another way to check packages security, so...

Ocramius:
> Moreover, as of SF4 + Flex, we might have another way to check packages security, so...

Which one?

Btw, this was already added in a couple other major skeleton apps, as it works very reliably (approach in this PR)

Ocramius:
> This will blocking an update that might fix some security vulnerabilities you have if it doesn't fix all of the vulnerabilities you have. Also the output is totally confusing, unless you know what is going on. I think this package is more likely to stop you upgrading than encourage you to upgrade, unless you are expert enough to know what it is.

Should business requirements override security (not endorsed, but happens a lot, sigh), you can remove the `roave/security-advisories` dependency anytime.

Pierstoval:
> Which one?

The same we already have in Symfony console with `security:check`? 👀

Ocramius:
@Pierstoval thanks, didn't see it :)

Pierstoval:
Actually, even if requiring `roave/security-advisories` is nice as a recommendation, it's the same kind of recommendation we have when we tell people to run `bin/console security:check` in CI or when deploying, this should be a recommendation only.

JarJak:
I think better solution would be something like https://github.com/symfony/symfony-standard/pull/954/files

fabpot:
Closing as this repo is now closed for feature addition. Flex is the future of Symfony.

PR #770, opened 2015-01-09
Fix requirements for the distribution bundle

gerryvdm:
Should still be ~3.0, with additional minimal requirement of 3.0.12, so we still can get updates to 3.1 etc.


fabpot:
Oops, I saw your comment but not the PR. I did the changes in all versions. Thanks.

PR #772, opened 2015-01-17
moved back dep binaries into vendor/dir

fabpot:
Move back all binaries from deps in `vendor/dir`, it's more standard amongst PHP projects using Composer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,4 @@
 /app/phpunit.xml
 /build/
 /vendor/
-/bin/
 /composer.phar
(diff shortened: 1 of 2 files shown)


gnugat:
Not to mention that it allows developers to create their own scripts in the `bin` folder, now that it is no longer ignored.

WouterJ:
:+1:

stof:
:+1:

hacfi:
:+1:

weaverryan:
:+1: And I think going into `master` is correct - it'll cause issues with documentation (hey, why do I get command not found when running `bin/XXXX`?) and I think doing it in 3.0 will help that a lot.

Pierstoval:
:+1: too, it's a bit confusing to have first "app/console", then "bin/console", but afterwards having console in the same dir as doctrine, because we never use doctrine "manually" , so a single sir for all vendor bin is great !

PR #773, opened 2015-01-17
Added a filter to check for local network ip range with php filter_var

aurimasniekis:
I am really hate then I use Vagrant / Docker dev environment every time I need to go to app_dev.php to add my host IP address to be able access application event then I am on local network IP range.

I think dev environment should be accessed from local network.

For checking if IP is in private network IP ranges I use PHP function `filter_var` and `FILTER_VALIDATE_IP` flag with `FILTER_FLAG_NO_PRIV_RANGE` which checks against IPv4 private network range defined in this [table](http://en.wikipedia.org/wiki/Private_network#Private_IPv4_address_spaces) but it fails validation for the IPv6 addresses starting with FD or FC. I also left the original `in_array` checking for specific IP's.

I also suggest to extract specific IP's array from if clause to separate array variable this will help then having real big list of IP's

### DIFF
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -11,7 +11,10 @@ use Symfony\Component\Debug\Debug;
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', 'fe80::1', '::1')) || php_sapi_name() === 'cli-server')
+    || !(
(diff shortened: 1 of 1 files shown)


stof:
The issue with this is that if you run a reverse proxy before your website in production, and then use a set of private IPs inside your own network, the website could see REMOTE_ADDR as being always a private IP (the private IP of your reverse proxy on your internal network), which would open the dev front controller to everyone.
So putting such condition in the file by default is a bad idea IMO (and IIRC, we already rejected the same proposal).

Note that doing it in your own project is fine if you are sure you won't end up in this case.

gcds:
Does not these line save from it `isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])` ?

Pierstoval:
In your case, maybe another solution is to delete the `app_dev.php` file from all VCS and prod server, and to use it only on your dev environment.

fabpot:
Closing for the reasons explained by @stof

PR #779, opened 2015-01-30
updated composer.json

fabpot:
I'd like to updated the composer.json for Symfony 2.3. Does it seem ok to everyone?


gnugat:
looks good to me

PR #782, opened 2015-02-17
Fixed default route path

florianv:
When using the Apache option `MultiViews` the path `/app` will be resolved to the `/app.php` file by Apache. So `/app/example` will resolve to `/app.php/example` and the application will respond with a 404 because the `/example` route does not exist.

This PR changes the example route to `/home` as it was also proposed in https://github.com/symfony/symfony-standard/issues/758.

### DIFF
--- a/src/AppBundle/Controller/DefaultController.php
+++ b/src/AppBundle/Controller/DefaultController.php
@@ -8,7 +8,7 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 class DefaultController extends Controller
 {
     /**
-     * @Route("/app/example", name="homepage")
+     * @Route("/home", name="homepage")
      */
     public function indexAction()
     {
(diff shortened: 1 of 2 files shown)


Tobion:
> /app/example will resolve to /app.php/example
Really? According to http://httpd.apache.org/docs/2.4/content-negotiation.html I would assume it would only look for files `/app/example.*` which does not exist. So it should not change the path.

florianv:
> Really? According to http://httpd.apache.org/docs/2.4/content-negotiation.html I would assume it would only look for files /app/example.* which does not exist. So it should not change the path.

Yes, it starts matching from the left and move to the next segment while the directory exists.

Tobion:
I see, because Apache also serves files at the beginning of a path, even when they have any path segment followed.

florianv:
> I see, because Apache also serves files at the beginning of a path, even when they have any path segment followed.

Yes, exactly. This PR is now superseded by this one https://github.com/symfony/symfony-standard/pull/784 but it can also be merged as it simplifies the path to the homepage

PR #784, opened 2015-02-17
Disabled MultiViews in .htaccess

florianv:
The `MultiViews` option enables Apache to negotiate and serve existing files depending on the request Accept header. For example `/config/test` will resolve to `/config.php/test` when accepting `text/html`.

So it prevents from using route paths beginning with the name of a file present in the `/web` folder. Moreover, this option is not needed because all requests that don't match an existing file will anyway pass through the Symfony front controller.

So this PR removes this option from the `.htaccess.`

> Note that https://github.com/symfony/symfony-standard/pull/782 can also be merged as it simplifies the route path.


web/.htaccess
```diff
@@ -5,6 +5,10 @@
 # to each configured DirectoryIndex file (e.g. index.php, index.html, index.pl).
 DirectoryIndex app.php
 
+# Disable the MultiViews option preventing from using route paths beginning with
+# "/app", "/app_dev" or "/config".
```
Tobion: `Disabling MultiViews prevents unwanted negotiation, e.g. "/app" should not resolve to the front controller "/app.php" but be rewritten to "/app.php/app"

web/.htaccess
```diff
@@ -5,6 +5,10 @@
...
+Options -MultiViews
```
Tobion: You probably need to test whether the module is enabled at all. `<IfModule mod_negotiation.c>`

web/.htaccess
florianv: Thanks, it's better

web/.htaccess
florianv: I nested it in the `<IfModule mod_rewrite.c>` too.

Tobion:
I would not nest it inside mod_rewrite. Even if mod_rewrite is not enabled, "/app" is not the same as "/app.php" in symfony terms.

florianv:
Thanks for reviewing @Tobion. Fixed

Tobion:
:+1:

PR #787, opened 2015-02-28
Running PHP-CS-fixer to get Symfony coding standards

msvrtan:
This are small changes to get Symfony standard edition to obey symfony coding standard (defined by PHP-CS-fixer).

### DIFF
--- a/app/SymfonyRequirements.php
+++ b/app/SymfonyRequirements.php
@@ -119,10 +119,10 @@ class PhpIniRequirement extends Requirement
      *
      * @param string        $cfgName           The configuration name used for ini_get()
      * @param bool|callback $evaluation        Either a boolean indicating whether the configuration should evaluate to true or false,
-                                               or a callback function receiving the configuration value as parameter to determine the fulfillment of the requirement
+     *                                         or a callback function receiving the configuration value as parameter to determine the fulfillment of the requirement
      * @param bool          $approveCfgAbsence If true the Requirement will be fulfilled even if the configuration option does not exist, i.e. ini_get() returns false.
-                                               This is helpful for abandoned configs in later PHP versions or configs of an optional extension, like Suhosin.
-                                               Example: You require a config to be true but PHP later removes this config and defaults it to true internally.
+     *                                         This is helpful for abandoned configs in later PHP versions or configs of an optional extension, like Suhosin.
+     *                                         Example: You require a config to be true but PHP later removes this config and defaults it to true internally.
      * @param string|null   $testMessage       The message for testing the requirement (when null and $evaluation is a boolean a default message is derived)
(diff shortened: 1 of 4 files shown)


hacfi:
@msvrtan Thanks for taking the time! Please create a separate PR for the changes of SymfonyRequirements.php at the repo https://github.com/sensiolabs/SensioDistributionBundle:
https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/skeleton/app/SymfonyRequirements.php

app/autoload.php
```diff
@@ -3,8 +3,8 @@
 use Doctrine\Common\Annotations\AnnotationRegistry;
 use Composer\Autoload\ClassLoader;
 
-/**
- * @var ClassLoader $loader
+/*
+ * @var ClassLoader
```
WouterJ: -1 for this change.

The old version was a completely valid doc block for the `$loader` variable. Seems like it isn't build into the PHP CS fixer yet.

The new version doesn't make any sense, as it's no longer a doc block and just a normal comment. No IDE will read it as being a docblock

app/check.php
```diff
@@ -21,7 +21,7 @@
 
 $messages = array();
 foreach ($symfonyRequirements->getRequirements() as $req) {
-    /** @var $req Requirement */
+    /* @var $req Requirement */
```
WouterJ: same here

app/SymfonyRequirements.php
```diff
@@ -119,10 +119,10 @@ class PhpIniRequirement extends Requirement
      *
      * @param string        $cfgName           The configuration name used for ini_get()
      * @param bool|callback $evaluation        Either a boolean indicating whether the configuration should evaluate to true or false,
-                                               or a callback function receiving the configuration value as parameter to determine the fulfillment of the requirement
```
Tobion: These changes must be done in SensioDistribution bundle as the header says.

app/autoload.php
GrahamCampbell: If you oneline this phpdoc, then php-cs-fixer won't break it.

app/check.php
GrahamCampbell: This regression has been fixed in php-cs-fixer now. It no longer makes this false change.

cordoval: i was actually liking this change :older_man:

GrahamCampbell: @cordoval this change is actually wrong. There is a difference between these two docblocks. The first one actually tokenises to something different to this one. This one is considered a comment, so all annotations have no meaning, where as, before it was considered a phpdoc comment, so annotations have meaning.

Tobion:
Closing as most changes are wrong.

PR #789, opened 2015-03-16
Fix yaml as @ is forbidden in YAML 1.2

jdespatis:
http://www.yaml.org/spec/1.2/spec.html#id2772075

(PR done on `2.5` branch, but this issue persists also in `2.6`, `2.7` and `master`)


fabpot:
Thank you @jdespatis.

PR #790, opened 2015-03-28
Send only headers in response to HEAD requests

icio:
I was surprised to find that this wasn't the case (though I wouldn't be surprised if there's a reason that I'm ignorant of). Per my testing, the routing component already treats HEAD requests like GET requests, so no issues there. I wanted to submit this patch in-case it was just something that had been overlooked and might find useful, or otherwise learn why it's a bad idea.

An alternative implementation might be to return only a HeaderResponse from the kernel when the request method is HEAD (where HeaderResponse is just a null-bodied Response), which would reduce duplication and increase testability. Perhaps also ought to use `Request::HTTP_HEAD`. Feedback welcome!

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -26,5 +26,10 @@ $kernel->loadClassCache();
 //Request::enableHttpMethodParameterOverride();
 $request = Request::createFromGlobals();
 $response = $kernel->handle($request);
-$response->send();
+
+if ($request->getMethod() == 'HEAD') {
+    $response->sendHeaders();
+} else {
+    $response->send();
+}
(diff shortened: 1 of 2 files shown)


pyrech:
IMO what you propose is not needed because it's already treated inside the `Response#prepare()` method ;) Have a look to it at https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Response.php#L298-305 to understand how it's done

icio:
Well! There you go. Thanks for sharing that with me, @pyrech

PR #792, opened 2015-04-02
updated deps

rpg600:
Actually user tests fails because the SecurityListener of the framework-extra-bundle still use the deprecated SecurityContext interface.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "incenteev/composer-parameter-handler": "~2.0"
     },
     "require-dev": {
-        "sensio/generator-bundle": "~2.3"
+        "sensio/generator-bundle": "~2.3",
+        "symfony/phpunit-bridge": "2.7.*@dev"
     },
     "scripts": {
         "post-root-package-install": [


Pierstoval:
:+1: Totally need this

stof:
FrameworkExtraBundle does not rely on the deprecated interface anymore in its latest release

rpg600:
https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/SecurityListener.php#L40

rpg600:
ok i see it from the tag v3.0.5 .. :)

stof:
@rpg600 but for Symfony 2.6+, it should pass ``null`` in this argument, and only inject the new services instead (it won't use the argument anyway). the fact that it still inject the deprecated service is a mistake in the implementation of the support of the old and new APIs in parallel

stof:
Actually no, there is no mistake. The argument is indeed removed in https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/51f3735d90ecb50d7277ce5102d5541725a2b1ca

PR #794, opened 2015-04-08
[config] add ide parameter.

aitboudad:
https://github.com/symfony/symfony-standard/pull/776#issuecomment-89763146

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -29,6 +29,7 @@ framework:
         handler_id:  ~
     fragments:       ~
     http_method_override: true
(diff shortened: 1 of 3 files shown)


Pierstoval:
I'm still not in favor of it :confused:

weaverryan:
@aitboudad Though completely logical, I think I'm against this too. It only applies to some people (though I am one of those people) and I *do* like showing how features are available to people. But it's also more clutter in an important file. I'd like to see less in general :).

fabpot:
I agree with @weaverryan as well. Closing it.

Issue #748, opened 2014-12-05
Doctrine DBAL 2.5 Breaks Symfony if DB Doesn't Exist


mattjanssen:
This is a problem with the most recent symfony-standard.

With the newest Doctrine DBAL 2.5, Doctrine introduced `Doctrine\DBAL\Connection::detectDatabasePlatform()` which is different than 2.4. Now in 2.5, any time a repository is instantiated (not even used) by the DI, we get a PDOException if the DB doesn't already exist.

Imagine trying to run `app/console doctrine:database:create` and getting a "Database doesn't exist!" error. Repositories as services are very common in most projects, so this is a huge issue.

To reproduce quickly, create a fresh Symfony project, letting composer install the AcmeDemoBundle. Use all default parameters. Now try to generate an entity. You'll get a PDOException. This is a new regression.
```
composer create-project symfony/framework-standard-edition symfony-standard
cd symfony-standard
app/console doctrine:generate:entity --entity=AcmeDemoBundle:Foo
```
You get error this on every OS/distro/platform:
```
  [Doctrine\DBAL\Exception\ConnectionException]
  An exception occured in driver: SQLSTATE[HY000] [1049] Unknown database 'symfony'

  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'symfony'

  [PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'symfony'
```

markitosgv:
is solved in DBAL 2.5.1?

stof:
no. Read the release announcement. They explicitly mention that they don't have an appropriate fix yet

markitosgv:
thanks @stof still waiting then

PuReWebDev:
Just wanted mention this issue still exists and the pull request associated with the above closed ticket did not get accepted, so the issue is still present. Thank you guys for opening the issue and bringing it to the attention of the appropriate devs. Found the page as I also was experiencing the same.

deudjuimw:
Hi all. Has the issue already been resolved or is there a workaround. Actually I'm following a tutorial with and facing the same error...

stof:
See https://github.com/doctrine/DoctrineBundle/issues/351#issuecomment-65771528 for the workaround

deudjuimw:
Ok, thank you !!!

Tobion:
Closing as the issue is resolved. I'm not getting that error anymore when using the latest versions.
[...]

PR #800, opened 2015-04-17
Update ORM requirements to allow DBAL 2.5 again

Tobion:
@Ocramius could doctrine release ORM 2.4.8 so https://github.com/doctrine/doctrine2/commit/e05930e71482df06982c332b8820a1ae51bd7f28 is tagged? I think orm <2.4.8 is incompatible with DBAL 2.5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,14 +9,14 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/symfony": "2.3.*",
-        "doctrine/orm": "~2.2.3||~2.3.0||~2.4.8||~2.5",
+        "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.2",
         "twig/extensions": "1.0.*",
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "~2.3",
-        "sensio/framework-extra-bundle": "~3.0,>=3.0.2",
+        "sensio/framework-extra-bundle": "^3.0.2",
         "sensio/generator-bundle": "~2.3",
         "incenteev/composer-parameter-handler": "~2.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -9,8 +9,7 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/symfony": "2.3.*",
-        "doctrine/orm": "~2.2,>=2.2.3,<2.5",
-        "doctrine/dbal": "<2.5",
+        "doctrine/orm": "~2.2.3||~2.3.0||~2.4.8||~2.5",
         "doctrine/doctrine-bundle": "~1.2",
         "twig/extensions": "1.0.*",
         "symfony/assetic-bundle": "~2.3",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "symfony/symfony": "2.3.*",
         "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.2",
-        "twig/extensions": "1.0.*",
+        "twig/extensions": "~1.0",
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",


composer.json
```diff
@@ -9,8 +9,7 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/symfony": "2.3.*",
-        "doctrine/orm": "~2.2,>=2.2.3,<2.5",
-        "doctrine/dbal": "<2.5",
+        "doctrine/orm": "~2.2.3||~2.3.0||~2.4.8||~2.5",
```
nicolas-grekas: single `|`?

Tobion: is the doc wrong https://getcomposer.org/doc/01-basic-usage.md#package-versions ?

stof: bith are valid as of alpha10. A single ``|`` works on older versions

stof: And I suggest dropping 2.2 and 2.3 here. They are EOLed versions of Doctrine

Tobion: changed. I also used the caret operator now. I guess it's time. And it's the responsibility of the people manually updating their dependencies anyway.

composer.json
```diff
@@ -9,15 +9,14 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/symfony": "2.3.*",
-        "doctrine/orm": "~2.2,>=2.2.3,<2.5",
-        "doctrine/dbal": "<2.5",
+        "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.2",
         "twig/extensions": "1.0.*",
```
stof: this one should be changed to ``~1.0`` to allow to use uptodate versions (the latest stable is 1.2.0)

Ocramius:
@Tobion ORM 2.4.7 works with DBAL 2.5.x

Tobion:
@Ocramius what do you mean with "works"? Does it not cause #748? How did you fix it for these version?

Ocramius:
It's not fixed in ORM 2.5 either.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

Tobion:
Then why does the error not appear anymore in ORM 2.5 if it's not fixed?

Ocramius:
Luck? I see no explicit fix happening.

That said, ORM 2.5 is DBAL 2.5 compatible, just platform auto-detection in DBAL is broken.

stof:
@Ocramius the auto-detection implementation in DBAL is fine: it is delayed until you retrieve the Platform from the connection (at which point the result of the detection is needed). As I explained several times already (including in http://www.doctrine-project.org/jira/browse/DBAL-1057 which tracks this officially), the issue is that the ORM was retrieving the platform in many places in the constructor instead of doing it only when it actually needs it. This means that just instantiating the ORM was asking for the platform at the right version.
The commit referenced above fixed this issue in 1 place in the ORM, and other places getting the platform in their constructor are the persisters, the hydrators and the DQL function parsers. None of them is instantiated until they are used IIRC, meaning that the requirement to be connected seems legitimate.
So I think it might be fixed by the commit referenced above.

Tobion:
Not yet. Doctrine needs to tag a new release which seems very hard to do. See https://github.com/doctrine/DoctrineBundle/issues/351#issuecomment-98545563

Ocramius:
s/hard/time-consuming-during-my-evening-that-is-already-fully-booked-for-these-weeks
:-)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

dbu:
ping @beberlei can you tag doctrine orm 2.4.8 maybe?

stof:
and also doctrine orm 2.5.1

Tobion:
Doctrine 2.4.8 has been released. So this can be merged.

fabpot:
Thank you @Tobion.

derrabus:
Don't forget to update the lock file. It's now out of sync and references DBAL 2.4.4 which is incompatible with the current requirements from composer.json.

GrahamCampbell:
The lock file gets updated before each release.

GrahamCampbell:
See https://github.com/symfony/symfony-standard/commit/a522d9de5d5d0eb10b4121d9c1d18f35c2ddbe46 for example.

GrahamCampbell:
It's not out of sync.

GrahamCampbell:
Just the newer branches not updated yet since there's not been a release there since.

derrabus:
@GrahamCampbell: I see, thanks for clarifying.

PR #802, opened 2015-05-01
Sandbox based on Docker Compose and official `nginx`, `mysql` and `php`

alexanderilyin:
PREREQUIREMENTS
=================

1. [Composer](https://getcomposer.org/download/) — It's required by Symfony2.
2. [Docker](https://docs.docker.com/installation/#installation) — It will run containers with preconfigured environments. 
3. [Docker Compose](https://docs.docker.com/compose/#installation-and-set-up) — It will link all containers to each other to make things works.
4. [boot2docker](http://boot2docker.io) — It will run docker if your host system is Mac or Windows.

QUICK START
===========

3 preconfigured containers based on official images. Start in 3 steps.

    php composer.phar install
    docker-compose up
    open http://`boot2docker ip`/config.php

If you prefer access to your site via `localhost` do the following:

    boot2docker ssh -vnNTL 8080:localhost:80 -L 8443:localhost:443
    open https://localhost:8443/app_dev.php


.gitignore
```diff
@@ -10,3 +10,5 @@
 /vendor/
 /bin/
 /composer.phar
+.idea
```
cordoval: remove these as these are dev specific

hacfi:
@alexanderilyin The code looks good but I think this should be placed in a separate repository. You could create a cookbook in the [symfony/symfony-docs](https://github.com/symfony/symfony-docs) repo instead.

cordoval:
does this work with boot2docker?

hacfi:
@cordoval I think it’s specifically created to work with boot2docker.

stof:
I agree with @hacfi here. The vagrant setup has already been rejected from the symfony-standard repo in the past

cordoval:
@alexanderilyin let this don't discourage you but work on a repo that has the docker files and share it in this thread, would rock to play with it.

.gitignore
alexanderilyin: @cordoval done.

alexanderilyin:
@hacfi possible you're right. I should write small article why it's useful with some cool screenshots like this:

screenshot 2015-05-01 00 03 20

@cordoval It was tested only in `boot2docker` on OS X but should work in any `docker` environment. Basically only `docker-compose.yml` should be included in project. But to make it work there are some configurations for nginx and some customization because of problems with `boot2docker` and permissions in shared folders.

Some details about used versions:
```
$ docker --version
Docker version 1.6.0, build 4749651
```
```
$ docker-compose --version
docker-compose 1.2.0
```

.gitignore
```diff
@@ -10,3 +10,4 @@
 /vendor/
 /bin/
 /composer.phar
+.mysql
```
alexanderilyin: This directory required to survive container restart and keep all data in MySQL.

fpm.conf
```diff
@@ -0,0 +1,20 @@
+[global]
+error_log = /proc/self/fd/2
+daemonize = no
+
+[www]
+# BOOT2DOCKER LIMITATIONS
+user = root
+group = root
+
```
alexanderilyin: `root` user required only if we want use official images under `boot2docker`.

mysql.cnf
```diff
@@ -0,0 +1,36 @@
...
+port		= 3306
+socket		= /var/run/mysqld/mysqld.sock
+
+[mysqld_safe]
+pid-file	= /var/run/mysqld/mysqld.pid
+socket		= /var/run/mysqld/mysqld.sock
+nice		= 0
+
+[mysqld]
+user		= root
```
alexanderilyin: `root` user required only if we want use official images under `boot2docker`.

nginx.conf
```diff
@@ -0,0 +1,71 @@
...
+
+    # pass the PHP scripts to FastCGI server from upstream phpfcgi
+    location ~ ^/(app|app_dev|config)\.php(/|$) {
+        fastcgi_pass phpfcgi;
+        fastcgi_split_path_info ^(.+\.php)(/.*)$;
+        include fastcgi_params;
+        fastcgi_param  SCRIPT_FILENAME $document_root$fastcgi_script_name;
+        fastcgi_param  HTTPS off;
+        # Allow access to /app_dev.php & /config.php
+        fastcgi_param  REMOTE_ADDR 127.0.0.1;
```
alexanderilyin: This is only to allow access `/config.php` & `/app_dev.php` without changes in Symfony sources.

cordoval:
i wonder if it would be useful at least a link to https://docs.docker.com/compose/ (btw i was using fig for my projects cc @alexkovalevych) would be useful and why it is more sanctioned than fig if it is?

fpm.conf
cordoval: i think these notes should be embedded in these files

alexanderilyin:
@cordoval answer is on http://www.fig.sh/index.html

```
Fig has been replaced by Docker Compose, and is now deprecated.
The new documentation is on the Docker website.
```
P. S.
I'll add details about prerequirements.

cordoval:
oh my bad true, had forgotten indeed, yes that is perfect

fabpot:
@alexanderilyin Thanks for your contribution. Unfortunately, we do not want to go down this road in this repository. The SE is only about the code you need to bootstrap a new Symfony project... and for any kind of Symfony projects. And your configurations makes some assumptions like a MySQL database or the usage of nginx. It would much better to create a standalone repo or contribute a cookbook for the docs. Thanks.

weaverryan:
@alexanderilyin I am interested in this kind of thing for the docs however - some unofficial (meaning community) repositories with Vagrant or Docker setups that we can inform people of in the official docs. See symfony/symfony-docs#4676 for an example. I think we're not doing a great service yet since this information is not included, and I'd be eager to discuss fixing that :).

Thanks!

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+phpfpm:
+  command: php-fpm --allow-to-run-as-root
+  image: php:5.6.8-fpm
+  volumes:
+    - fpm.conf:/usr/local/etc/php-fpm.conf
```
kayue: @alexanderilyln I got `file exists at /mnt/sda1/var/lib/docker/aufs/mnt/eb13125c1167ab4a391994d402ac8181b13901875d2c1350793452242ca275b2/usr/local/etc/php-fpm.conf, can't create volume there` error when `docker-compose up`. Same to other NGINX config below.

I am using boot2docker, am I missing anything?

kayue: @alexanderilyin FYI if I mount a directory, instead of a file, then it does build.

mmi-rperez:
@alexanderilyin is this project death or alive? are you following up?

PR #813, opened 2015-05-22
Changed ORM default naming strategy to underscore

hason:
This PR partially fixes the problem with case-sensitive identifiers http://dev.mysql.com/doc/refman/5.0/en/identifier-case-sensitivity.html

Replaces https://github.com/doctrine/DoctrineBundle/pull/406

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -68,6 +68,7 @@ doctrine:
 
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
+        naming_strategy: doctrine.orm.naming_strategy.underscore
         auto_mapping: true
 
 # Swiftmailer Configuration


Pierstoval:
I think that this can be very useful :+1:

hacfi:
:+1:

TomasVotruba:
:+1:

stof:
:+1: The default value cannot be changed in DoctrineBundle because it would be a BC break for existing projects, but the underscore is a much saner choice than the default naming strategy keeping camel-cased names. It has much less compatibility issues across platforms.

fabpot:
Thank you @hason.

weaverryan:
:+1: Wish we had done this awhile ago :)

stof:
@weaverryan better late than never

PR #815, opened 2015-05-31
[CRITICAL] PSR-4 path to `SymfonyStandard` is incorrect

zerkms:
It was simply replaced at 6272b3388862e57f6fd688f06a1d1c9896fd4ffd `psr-0` with `psr-4` but it additionally requires fixing the path to the namespace.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "type": "project",
     "description": "The \"Symfony Standard Edition\" distribution",
     "autoload": {
-        "psr-4": { "": "src/", "SymfonyStandard\\": "app/" }
+        "psr-4": { "": "src/", "SymfonyStandard\\": "app/SymfonyStandard/" }
     },
     "require": {
         "php": ">=5.3.3",


Tobion:
:+1:

gremo:
Do this affect only the new directory structure as @zerkms explained [here](http://stackoverflow.com/questions/30555986/how-can-i-use-the-new-directory-structure-with-symfony-2-7)?

fabpot:
Thank you @zerkms.

zerkms:
@gremo it affects creating a new directory structure and a demo `Acme` bundle

peterrehm:
@fabpot This should get into a release as you get the `Class SymfonyStandard\Composer is not autoloadable, can not call post-root-package-install script` when running the install script by composer.

````php
    composer create-project symfony/framework-standard-edition my_project_name "2.7.*"

Pierstoval:
@peterrehm Seems this problem can be solved if you use the [symfony-installer](https://github.com/symfony/symfony-installer)?

peterrehm:
Should be the case, however the composer install should be still supported as long as we have it documented.

andrey1s:
@Pierstoval I use the directory structure Symfony 3.0 that is not supported `symfony-installer`

peterrehm:
This should help for now:

````
composer create-project symfony/framework-standard-edition my_project_name "2.7.x-dev*"

GrahamCampbell:
`2.7.*@dev` would be the correct thing here.

weaverryan:
This should be in the new 2.7.1 tag now.

PR #826, opened 2015-06-19
adding light documentation to security.yml

weaverryan:
Hi guys!

When we removed AcmeDemoBundle recently, we lost a lot of "example" security configuration that was there before. I realized this in a training - you open this file, but don't know where to go next. I've added some very basic links to help get the user moving.

@javiereguiluz does this make sense to you?

Thanks!

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -1,13 +1,24 @@
+# To get started with security, check out the documentation:
+# http://symfony.com/doc/current/book/security.html
 security:
 
+    # http://symfony.com/doc/current/book/security.html#where-do-users-come-from-user-providers
     providers:
         in_memory:
             memory: ~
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@weaverryan yes it makes sense to me. In a previous issue, @Pierstoval asked if we could add help notes in this file and I suggested to not do that. But I guess there is a sweet spot between adding nothing and adding a lot of documentation. I think your pull request is exactly that sweet spot. Thanks!

weaverryan:
@javiereguiluz I agree - we don't want to maintain documentation on here. Thanks!

Pierstoval:
:+1:

fabpot:
Thank you @weaverryan.

PR #834, opened 2015-07-22
Added a "Welcome Page" for new Symfony apps

javiereguiluz:
### Problem

When you create a new Symfony app (`symfony new my_project`) and follow the installer instructions, when browsing the application for the first time, you'll get a 404 error:

old_welcome_screen

### Solution

In this PR we add a new simple welcome page to avoid that error. This page is designed for newcomers, so we'll guide them exactly to where they should go: to the doc explaining how to create their first page with Symfony:

new_welcome_screen_ok


app/Resources/views/default/index.html.twig
```diff
@@ -1,5 +1,94 @@
...
+                    </p>
+                </div>
+
+            {% endif %}
+        </div>
+    </div>
+{% endblock %}
+
+{% block stylesheets %}
+<style type="text/css">
```
WouterJ: type attribute is redundant and not required

Pierstoval:
I don't think that writing the whole logic of the welcome page directly in the `DefaultController` is a good idea... Isn't it possible to add this behavior in another controller from the framework and do a simple `forward`? 
 
We removed the entire `AcmeDemoBundle` a little because of the fact that installing a new Symfony app implied removing this bundle manually, so adding code again that would have to be removed by the devs seems quite a "rewind back"...?

src/AppBundle/Controller/DefaultController.php
```diff
@@ -9,10 +9,15 @@
 class DefaultController extends Controller
 {
     /**
-     * @Route("/app/example", name="homepage")
+     * @Route("/", name="homepage")
      */
     public function indexAction()
     {
-        return $this->render('default/index.html.twig');
+        require_once __DIR__.'/../../../app/SymfonyRequirements.php';
```
weaverryan: Add a comment about this, something like:

```php
// replace this example code with whatever you need
```

weaverryan:
@Pierstoval The logic is just a couple of lines of code - that's quite easy to remove.

:+1: 

@javiereguiluz is already aware of this, but we'll need a docs PR when this is merged - it requires some screenshots, so there's no use doing it before a merge.

javiereguiluz:
I've done the suggested changes.

Given that the current situation is dreadful (the first thing you see about Symfony is an error page) I ask you to please make a decision about this PR as quickly as possible. Thanks!

lyrixx:
:+1:

pgodel:
Great improvement!

:+1:

ogizanagi:
:+1:

src/AppBundle/Tests/Controller/DefaultControllerTest.php
```diff
@@ -12,6 +12,7 @@ public function testIndex()
 
         $crawler = $client->request('GET', '/');
 
-        $this->assertTrue($crawler->filter('html:contains("Homepage")')->count() > 0);
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
+        $this->assertContains('Welcome to Symfony', $client->getResponse()->getContent());
```
stof: I suggest keeping a crawler-based example too in the test (or remove the unused ``$crawler`` variable)

TomasVotruba:
:+1:

hason:
:+1:

src/AppBundle/Tests/Controller/DefaultControllerTest.php
javiereguiluz: Good point! I've changed it by:

```php
$this->assertContains('Welcome to Symfony', $crawler->filter('#container h1')->text());
```

Thanks.

romqin:
:+1:

app/Resources/views/default/index.html.twig
```diff
@@ -1,5 +1,94 @@
...
+                                c-1.1-0.7-2.3-0.5-3.4,0.3c-0.5,0.4-0.8,1-1,1.6c-0.4,1.5,0.4,2.9,0.8,3.4l0.9,1c0.2,0.2,0.6,0.8,0.4,1.5c-0.3,0.8-1.2,1.3-2.1,1
+                                c-0.4-0.2-1-0.5-0.9-0.9c0.1-0.2,0.2-0.3,0.3-0.5s0.1-0.3,0.1-0.3c0.2-0.6-0.1-1.4-0.7-1.6c-0.6-0.2-1.2,0-1.3,0.8
+                                C4.3,38.4,4.7,40,6.8,40.8z M46.1,20.9c0-4.2-3.2-7.5-7.1-7.5h-3.8C34.8,10.8,32.7,9,30.2,9L-2.3,9.1c-2.8,0.1-4.9,2.4-4.9,5.4
+                                L-7,58.6c0,4.8,8.1,13.9,11.6,14.1l34.7-0.1c3.9,0,7-3.4,7-7.6L46.1,20.9z M-0.3,36.4c0-8.6,6.5-15.6,14.5-15.6
+                                c8,0,14.5,7,14.5,15.6S22.1,52,14.2,52C6.1,52-0.3,45-0.3,36.4z M42.1,65.1c0,1.8-1.5,3.1-3.1,3.1H4.6c-0.7,0-3-1.8-4.5-4.4h30.4
+                                c2.8,0,5-2.4,5-5.4V17.9h3.7c1.6,0,2.9,1.4,2.9,3.1V65.1L42.1,65.1z"/>
+                        </svg>
+
+                        Read Symfony documentation to learn
+                        <a href="http://symfony.com/doc/{{ symfony_version }}/book/page_creation.html">How to create your first page in Symfony</a>
```
WouterJ: why not just use `constant('\\Symfony\\Component\\HttpKernel\\Kernel::VERSION')` here as well? Symfony.com is capable of handling x.y.z versions in the URL.

app/Resources/views/default/index.html.twig
```diff
@@ -1,5 +1,94 @@
 {% extends 'base.html.twig' %}
 
 {% block body %}
-    Homepage.
+    <div id="wrapper">
+        <div id="container" class="fade-in">
+            <div id="welcome">
+                <h1><span>Welcome to</span> Symfony {{ constant('\\Symfony\\Component\\HttpKernel\\Kernel::VERSION') }}</h1>
```
WouterJ: is there really a need for double `\` ? (and a `\\` prefix?)

javiereguiluz: The double slash is mandatory, but the leading one not. Changed. Thanks.

app/Resources/views/default/index.html.twig
javiereguiluz: Thanks for the tip. I've simplified the template. Thanks.

WouterJ: I just thought of a case that is no longer supported: unstable versions (where the version will be `2.8-dev` for instance of `2.8-rc1`).

trq:
This is a simple but very effective improvement in my opinion. Great idea @javiereguiluz

app/Resources/views/default/index.html.twig
GrahamCampbell: Also, you shouldn't use a leading slash in the class name. Class names are defined without that slash.

javiereguiluz: @GrahamCampbell removed the leading slash. Thanks!

@WouterJ what about making this check for unstable versions?

```twig
{{ symfony_version|length > 3 ? 'current' : symfony_version }}
```

gnugat:
I think we're trying to mix two different concepts here:

* providing a skeleton allowing developers to start writing their application right away
* checking if the environment is correctly set up

When I use the standard edition, I don't directly open a page in the browser: I first start writing my app and then I'll check it later.
AcmeDemoBundle was removed because its purpose was to showcase the code of a Symfony application, and having it in the standard edition just slowed down developers who actually want to just start coding.

On the other hand, I do think we're missing a good tool to check if the environment is set up correctly. But I'd rather see it as a standalone script you could run everywhere.

javiereguiluz:
For people using the Symfony Installer, all technical requirements are checked during the installation and you can see the list of things to fix. In case you ignore this list, the first thing you'll see in the browser is the message telling you that some requirements aren't met.

For people still using Composer to install Symfony, they don't get the nice requirements checker of the installer, so it's nice to see the welcome page reminding them about these errors.

For Symfony experts, the new welcome page is just a 3 lines controller and 1 Twig template. They will clean everything very easily.

cordoval:
Contrary to the common denominator I am :-1: because this creates confusion, maybe even documented confusion. I would have initially thought that a default route on / is more than enough.

javiereguiluz:
@cordoval I don't understand your comment :confused: This PR is precisely to provide a default route to `/`, which currently is missing.

cordoval:
Instead of the demo doing a /examples/page, it could just be /. The demo in the first place missed the point completely.

app/Resources/views/default/index.html.twig
WouterJ: @javiereguiluz just thought of a nicer solution: `{{ symfony_version[:3] }}`. If the version is `2.7.3`, it'll link to `/doc/2.7/book/...` and when the version is `3.0-dev`, it'll link to `/doc/3.0/book/...`, which is redirected to `/doc/master/book/...`.

Btw, can you please add the 2.8 version to the documentation?

WouterJ:
I have a bit the same feeling as @cordoval and @gnugat. I would love to keep these things simple, as it now contains quite a bit of very confusing things (e.g. require_once in the controller, data urls, lots of logic).

What about removing the requirement check and just link to `/web/check.php`?

cordoval:
My feeling is broader actually, I remember those days in which symfony was simple :blush: and that we were happy escaping the magic of sf14 and give freedom to users. I know it is a hard critic, I did one project in sf14 for an old client recently and it felt to me simpler than working with symfony2. That feeling sank in me heavily.

WouterJ:
I have the feeling that your comments completely not related to this PR, @cordoval. Can you maybe create a new issue for your thoughts on this (as it can lead to interesting results, when thinking about what can be made easier?) in the symfony repository and avoid the noice here? Thanks! :smiley:

javiereguiluz:
@WouterJ I like your last proposal. I've simplified everything in 3196523. Thanks.

Pierstoval:
@cordoval I don't understand why things are "more complicated" with this PR, even if I'm :-1: for the way it was implemented.
 
For the rest, I still think that, even if its "only 3 lines of code and a template", it's still much more than "one single line of code" for me...

WouterJ:
:+1: for the current state and merging this one quickly

fabpot:
@javiereguiluz Can you update the screenshot in the description to avoid comments about the previous version?

javiereguiluz:
Done.

fabpot:
:+1:

weaverryan:
:+1:

sstok:
It's not magical or anything (like that it's part of the FrameworkBundle defining a default route), its a sensible default with a single page (not a complete bundle). And you are going to change (rather then delete - aka more work) these files anyway, so :+1: for me.

@cordoval 

> My feeling is broader actually, I remember those days in which symfony was simple :blush: and that we were happy escaping the magic of sf14 and give freedom to users. I know it is a hard critic, I did one project in sf14 for an old client recently and it felt to me simpler than working with symfony2. That feeling sank in me heavily.

Never used Symfony 1.4 :) so can't really tell, *what is simple is not automatically easy*.
I understand your point, simple implies (simplistic or minimalistic) while easy describes (easy to use and change). I agree on the 'its starts directly from the box' vs 'clean from the box', that it's better (simpler) to get started directly without having to remove allot of demo code. But we also want to keep the experience to be easy and pleasant for newcomers.

src/AppBundle/Controller/DefaultController.php
```diff
@@ -4,15 +4,17 @@
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
-use Symfony\Component\HttpFoundation\Request;
```
fabpot: We decided at some point to keep such a use statement to make it easier for people to start working right away. Not sure if we need to keep it now.

app/Resources/views/default/index.html.twig
```diff
@@ -1,5 +1,81 @@
 {% extends 'base.html.twig' %}
 
 {% block body %}
-    Homepage.
+    {% set symfony_version = constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION') %}
```
fabpot: Small suggestion: I would inline this instead of creating a var

gnugat:
> @javiereguiluz
> For people using the Symfony Installer, all technical requirements are checked during the installation and you can see the list of things to fix. In case you ignore this list, the first thing you'll see in the browser is the message telling you that some requirements aren't met.
>
> For people still using Composer to install Symfony, they don't get the nice requirements checker of the installer, so it's nice to see the welcome page reminding them about these errors.
>
> For Symfony experts, the new welcome page is just a 3 lines controller and 1 Twig template. They will clean everything very easily.

I wouldn't use the installer nor composer bootstrap when creating a new server for an existing sf app. I'm actually longing for a standalone script that could be used anywhere, but that's another topic, sorry for bringing it here.

Now back to the actual PR: I don't get the point of having a default page when all I want is an empty skeleton.

src/AppBundle/Controller/DefaultController.php
sstok: If we keep it, it should at least be in the `indexAction` method to show why its here.

weaverryan:
@gnugat I helped a user on Twitter just last week - they got the 404 page and thought that *tons* of things must be broken with their setup because they got an error page when first loading things up. That's the point.

gnugat:
@weaverryan in the case of developers trying out Symfony for the first time, doesn't it make sense to provide them a different application (let's say the AcmeDemo show case application)? To me it sounds more like a lack of explanation / choice on the [download page](http://symfony.com/download).

pgodel:
Guys, we are trying to help the newcomers, if you already know Symfony, will know that you need to delete 3 lines of code and one twig template, and that won't take much time even if you have to do it multiple times.

Pierstoval:
@gnugat It depends of what kind of "newcomer" you are. In my case, when I want to discover new frameworks/languages, I won't look at the code at first. I'll create an app. So in the case of Symfony, if I had to discover it today, I'd create an app and start to code right now by following the docs/books. 
 
Even if the Symfony Demo app is great in many points, it depends on each person, and IMO is more for presentations and learnings.
 
And many other developers will give you different answers. That's why providing a very small **working** skeleton is absolutely important.

fabpot:
Thank you @javiereguiluz.

yosymfony:
Good

PR #839, opened 2015-07-23
update Doctrine Configuration in config.yml

ykseiljon:
allow to use other database driver from parameters.yml

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -52,7 +52,7 @@ assetic:
 # Doctrine Configuration
 doctrine:
     dbal:
-        driver:   pdo_mysql
+        driver:   "%database_driver%"
         host:     "%database_host%"
         port:     "%database_port%"
         dbname:   "%database_name%"


ogizanagi:
The `%database_driver%` does not exist in `app/config/parameters.yml.dist`

ykseiljon:
I updated app/config/parameters.yml, then i need to change config.yml to connect to other database driver and worked well.
If I use assistant to configure database connection it show that parameter (database_driver on app/config/parameters.yml).

ogizanagi:
The issue is that you updated the app configuration to use the `database_driver`, but you didn't edit the `parameters.yml.dist` file from this repository within your PR. 
`parameters.yml.dist` file which is used by the Composer Incenteev parameter handler to ask the user the parameters values after a `composer install`.

Doing a fresh install of symfony standard edition will then fail, because of that missing parameter from the `.dist` file. The `database_driver` parameter will never be created in `parameters.yml`.

Also, the `database_driver` used to exists, but has been removed. (See above referenced PRs for more informations about that)

ykseiljon:
thanks, after check it i understand reason, it work good if i write driver of appropriate database in config.yml

Pierstoval:
IIRC, the `database_driver` was removed from `parameters.yml` because it's mostly a "project-related" parameter, not a "platform-related". 
 
Everything in the `parameters.yml` should depend only on the platform in which you deploy/dev/test your app, and the rest should stay in `config.yml`, this is why this parameter was removed from the dist file.

xabbuh:
@Pierstoval is right. That's the reason why the parameter was removed in #776 (see https://github.com/symfony/symfony-standard/pull/776#discussion_r23593282). Therefore I'm going to close here.

ykseiljon:
I get it.  But I think is necessary to update configuration assistant, because it still write database_driver in parameter.yml. Thanks for answer.

xabbuh:
Did you experience that behaviour with a freshly installed Symfony application?

Pierstoval:
@xabbuh I've tested on a freshly installed 2.7 version and actually, the `SensioDistribution:Configurator` controller adds the `database_driver` to `parameters.yml`.
 
I didn't use this component for months, but actually, it also generates a completely inline parameters.yml file. I'm on MacOSX10.10.

ykseiljon:
@xabbuh, @Pierstoval is right. And I appreciate the explanations of the change reasons and understood. It's sense to me too.

Pierstoval:
The issue is then related to `SensioDistributionBundle`, I think we can close it here

xabbuh:
see sensiolabs/SensioDistributionBundle#219

PR #844, opened 2015-08-13
updated composer.json for SensioGeneratorBundle

OskarStark:
this is more readable, as the 2.5.2 was anyway locked

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "incenteev/composer-parameter-handler": "~2.0"
     },
     "require-dev": {
-        "sensio/generator-bundle": "~2.3"
+        "sensio/generator-bundle": "~2.5"
     },
     "scripts": {
         "post-root-package-install": [
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "73f15ad91f600008506b4294b7b7d01c",
+    "hash": "66bc04cc32aab79c2ed1413b5e92bc02",
     "packages": [
         {
             "name": "doctrine/annotations",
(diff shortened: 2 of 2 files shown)


Tobion:
What is wrong with the 2.3 version? If it works, the requirement should not be increased.

OskarStark:
there is nothing wrong, but the composer.lock file now pins 2.5.2 anyway...

https://github.com/symfony/symfony-standard/blob/2.8/composer.lock#L1568

stof:
@OskarStark and what is the issue ? The composer.lock contains the version selected by composer. This does not mean we have to bump the min requirement

OskarStark:
there is no issue.

IMO it is not very transparent that ```~2.3``` uses ```2.5.2``` (just by reading the composer.json)

i know this is locked in the composer.lock, but didn't you think if we write ```~2.5``` is the same result with a much more clearer info to a developer?

this is just a proposal, feel free to close this PR ;-)

Pierstoval:
I think @fabpot often updates dependencies on the standard edition so the
next release has an updated composer.lock

Anyway we should still keep "lowest" dependencies in composer.json to avoid
conflicts with third-party bundles that may require an older version

2015-08-14 10:21 GMT+02:00 Oskar Stark <notifications@github.com>:

OskarStark:
to be more clear: in this case we could also use ```~2.0```, right?

my recommendation is, if we have patch versions, i would use them in the json file

OskarStark:
@Pierstoval in this case, this would not change anything at all for conflicts, as the 2.5.2 is already locked....

Pierstoval:
Yep it is locked, but when you run a `composer update` for all deps and a package requires an older version, then the good one can be installed. But it's quite rare...

OskarStark:
@Pierstoval , ok you are right, but is this really needed for this bundle?

as you said, @fabpot often updated some patch versions for different releases, but never the 2.3 of GeneratorBundle ;-)

Pierstoval:
I think that patch versions should be changed whenever changes in the main project (we're on the standard edition here, fully customizable, it's not like it was the framework repo) include modifications that are incompatible with older versions of the projects, it's more a matter of BC breaks policy explanation, saying something like "SensioGeneratorBundle 2.3+ is compatible", even if the lock file is on 2.5

Tobion:
In the composer.json file you should put the MINIMUM viable version. What version is then locked or chosen by the user, is a different thing.

PR #846, opened 2015-08-27
Remove .gitkeep and create .gitignore file

jbdelhommeau:
As explained here .gitkeep,  is not a feature of Git. He was replaced by .gitignore.

http://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep?answertab=active#tab-top

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,6 @@
 /app/cache/*
 /app/config/parameters.yml
 /app/logs/*
-!app/cache/.gitkeep
-!app/logs/.gitkeep
 /app/phpunit.xml
 /build/
 /vendor/
--- a/app/cache/.gitignore
+++ b/app/cache/.gitignore
@@ -1,0 +1,1 @@
+!.gitignore
--- a/app/logs/.gitignore
+++ b/app/logs/.gitignore
@@ -1,0 +1,1 @@
+!.gitignore


weaverryan:
Hey @jbdelhommeau! In the link, it doesn't say that `.gitignore` is more valid than `.gitkeep` - just that there are 2 competing conventions. Or is there some additional agreement with the `.gitignore` method?

Pierstoval:
Actually, the `.gitignore` file you added in replacement of `.gitkeep` is incorrect.
 
The contents have to be the following:

```ini
*
!.gitignore
```

It's mandatory to add the `!.gitignore` part, else the ignore file will... Ignore himself, so the ignore directives won't be taken in account because the ignore file is ignored (kind of a paradigm loop, anyway).
 
I don't know if it's the best solution, but actually it can work.

weaverryan:
It works, but if anything, it just looks like a more technical (complex) way of getting to the same solution (but it is certainly valid). So unless this proposed way is really a better or more standard way, I'll say :thumbsdown:

hacfi:
Personally I prefer the current way: one `.gitignore` which contains all the rules - much easier to overview what is going on.

lyrixx:
More over, now each time you will run `rm -rf app/cache`, git working dir will be dirty

weaverryan:
Ok, thanks for the suggestion, it's valid, but I'm closing this

PR #850, opened 2015-09-16
Update SE for Symfony 3.0

fabpot:


gnugat:
Test classes could be autoloaded only in a development environment by moving them from `src/AppBundle/Tests` to `tests` and adding a `autoload-dev` section in composer.
What do you think about this suggestion?

stof:
I agree with @gnugat here. Separating tests from the runtime source is a good idea IMO

.gitignore
```diff
@@ -1,11 +1,10 @@
-/web/bundles/
 /app/bootstrap.php.cache
-/app/cache/*
 /app/config/parameters.yml
-/app/logs/*
-!app/cache/.gitkeep
-!app/logs/.gitkeep
 /app/phpunit.xml
```
stof: should be changed too as we moved the phpunit config at the root

stof:
this also requires an update of the HttpKernel component to change the location of the cache and logs

weaverryan:
@gnugat @stof @fabpot is just making the structure match what was decided in #584 - so we should focus only on that (and open a new issue about tests for discussion).

Tobion:
Created #855 for tests

fabpot:
Let's move forward, merging now.

fabpot:
Needs more work

.gitignore
```diff
@@ -1,11 +1,10 @@
-/web/bundles/
 /app/bootstrap.php.cache
```
hacfi: this is moved to `/var/` as well

phpunit.xml.dist
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
+         backupGlobals="false"
+         colors="true"
+         bootstrap="app/bootstrap.php.cache"
```
hacfi: `bootstrap="var/bootstrap.php.cache"`

var/bootstrap.php.cache
```diff
@@ -0,0 +1,3078 @@
+<?php
```
hacfi: No need to commit this

Tobion:
I'd recommend to merge #854 first.

bin/console
```diff
@@ -7,8 +7,8 @@
 
 set_time_limit(0);
 
-require_once __DIR__.'/bootstrap.php.cache';
-require_once __DIR__.'/AppKernel.php';
+require_once dirname(__DIR__).'/app/bootstrap.php.cache';
```
weaverryan: As others have said, this lives in `/var` - and the same change will need to go in the front controllers.

fabpot:
Besides the new directory structure, I've made another change that should help: sessions are now stored outside of the `cache/` directory, and directly into the `var/` dir.

Pierstoval:
@fabpot Why is it better to store the sessions in `var` instead of `cache` ?

fabpot:
@Pierstoval Because clearing the session store should be decoupled from clearing the Symfony cache, the two are not related. And when using `cache:clear`, all the cache is flushed. Also, storing sessions in the cache has some problems with Windows (see symfony/symfony#15862 -- not confirmed though)

fabpot:
This one is ready for a last review before merge.

bin/symfony_requirements
```diff
@@ -1,6 +1,7 @@
+#!/usr/bin/env php
 <?php
```
yguedidi: Are you sure about the file permissions change?

michaelcullum:
Should `var/sessions/` not also have a `.gitkeep` and be added to the `.gitignore`?

fabpot:
@michaelcullum right, fixed

PR #852, opened 2015-09-16
Use a proper version constraint

GrahamCampbell:
Any reason why this wasn't used already? I noticed this when looking at https://github.com/symfony/symfony-standard/pull/850.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     },
     "require": {
         "php": ">=5.3.9",
-        "symfony/symfony": "2.8.x-dev",
+        "symfony/symfony": "2.8.*@dev",
         "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.4",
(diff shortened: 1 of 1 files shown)


gnugat:
You could push it a little bit further: `^2.8.0@dev`

mickaelandrieu:
humm, should be also done for Symfony3

GrahamCampbell:
> You could push it a little bit further: ^2.8.0@dev

:-1: because all other branches are like `2.7.*`.

GrahamCampbell:
`^2.8.0@dev` - that's wrong anyway, because it would include 2.9+ too. Yes, I know 2.8 is the last 2.x minor release, but that version constraint would be wrong, semantically.

GrahamCampbell:
Ping @fabpot. :)

Tobion:
Thank you @GrahamCampbell.

GrahamCampbell:
:)

Issue #788, opened 2015-03-14
Using 'app/autoload.php' instead of 'app/bootstrap.php.cache' in app_dev.php by defaults


zomberg:
Symfony's documentation has a great article http://symfony.com/doc/current/cookbook/debugging.html.
Using 'app/autoload.php' instead of 'app/bootstrap.php.cache' in app_dev.php is really help with debug application.

When I go with debugger into 'bootstrap.php.cache' the debugging become more complicated.

On my work and my personal projects I always disable bootstrap.php.cache.
All my colleagues which used Symfony 2 also pointed that started new projects with disabling bootstrap.php.cache and enable app/autoload.php instead.
I see that this approach also used in https://github.com/gimler/symfony-rest-edition, which promoted on http://symfony.com/distributions page.

What you think about using app/autoload.php by default in app_dev.php?

weaverryan:
Hey @zomberg!

The big reason against this will be that Fabien has always wanted the different environments to have as little difference as possible, to avoid issues that happen in one, but not the other. And I *do* think that's important. Of course, we *do* promote exactly what you're saying here in the docs: http://symfony.com/doc/current/cookbook/debugging.html.

One idea I can think of is including the require line for `app/autoload.php` in `app_dev.php`, but commented out, and with 1 commented line explaining how you can improve debugging.

Cheers!

zomberg:
Hi Ryan!

Ok, I undestand your point of view.

I think it is good idea to include in this code in comments with link to http://symfony.com/doc/current/cookbook/debugging.html page.

Thanks.

weaverryan:
:+1: to including it in the code with comments!

jrobeson:
alternatively.. we could pull the autoloader out of the bootstrap.php.cache file  so then you simply comment include 'bootstrap.php.cache'; if you don't want the cache.

Tobion:
@jrobeson this sounds like a good plan both from a usuability and from an architecture perspective as there is not reason to mix the bootstrap with the already existring autoload.php file.

stof:
looks like a good idea indeed (we should still keep it in bootstrap.php.cache for now for BC reasons though)

sstok:
I actually solved this with a condition for the app file (using Sf3 directory structure btw).
And for the PHPUnit I don't use the bootstrap.php.cache but rather the autoloader directly.

```
// Don't use the class caches when xdebug is enabled
// so its easier to debug errors.
[...]

PR #854, opened 2015-09-17
extract autoloading out from bootstrap

Tobion:
Fixes https://github.com/symfony/symfony-standard/issues/788#issuecomment-98993898 and https://github.com/symfony/symfony/pull/14776#issuecomment-106862443. 

- by default class aggregation is still used as it is probably beneficial by default for most people (no opcache at all or stats enabled)
- People that do not want to use the aggregation, can simply comment the line. They do not have to rewrite code. E.g. when using opcache with `opcache.validate_timestamps=0` in production there is no point in using the bootstrap.
- No debugging hell anymore as the class aggregation is only done in in no-debug mode
- All environments still use the same code path, i.e. load autoload first. The only difference is the inlusion of the aggregation.
- This way we can also simplify the bootstrap generation in the distribution bundle in the next major version of it.
- several imrprovements like using `require` instead of `require_once` for the autoloader to rely on the return value and `include_once` instead of `require_once` for the bootstrap as the app is also working without it

### DIFF
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -3,11 +3,15 @@
 use Doctrine\Common\Annotations\AnnotationRegistry;
 use Composer\Autoload\ClassLoader;
 
+error_reporting(error_reporting() & ~E_USER_DEPRECATED);
+
 /**
  * @var ClassLoader $loader
  */
 $loader = require __DIR__.'/../vendor/autoload.php';
 
 AnnotationRegistry::registerLoader(array($loader, 'loadClass'));
+
+include_once __DIR__.'/bootstrap.php.cache';
 
 return $loader;
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -12,6 +12,4 @@ $loader = require __DIR__.'/../vendor/autoload.php';
 
 AnnotationRegistry::registerLoader(array($loader, 'loadClass'));
 
-include_once __DIR__.'/bootstrap.php.cache';
-
 return $loader;
--- a/app/console
+++ b/app/console
@@ -7,7 +7,7 @@
 
 set_time_limit(0);
 
-require_once __DIR__.'/bootstrap.php.cache';
+require_once __DIR__.'/autoload.php';
 require_once __DIR__.'/AppKernel.php';
 
 use Symfony\Bundle\FrameworkBundle\Console\Application;
--- a/app/console
+++ b/app/console
@@ -1,18 +1,21 @@
 #!/usr/bin/env php
 <?php
 
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Debug\Debug;
+
 // if you don't want to setup permissions the proper way, just uncomment the following PHP line
 // read http://symfony.com/doc/current/book/installation.html#configuration-and-setup for more information
 //umask(0000);
 
 set_time_limit(0);
 
-require_once __DIR__.'/autoload.php';
+/**
+ * @var Composer\Autoload\ClassLoader $loader
+ */
+$loader = require __DIR__.'/../app/autoload.php';
 require_once __DIR__.'/AppKernel.php';
-
-use Symfony\Bundle\FrameworkBundle\Console\Application;
-use Symfony\Component\Console\Input\ArgvInput;
-use Symfony\Component\Debug\Debug;
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
@@ -20,6 +23,8 @@ $debug = getenv('SYMFONY_DEBUG') !== '0' && !$input->hasParameterOption(array('-
 
 if ($debug) {
     Debug::enable();
+} else {
+    include_once __DIR__.'/bootstrap.php.cache';
 }
 
 $kernel = new AppKernel($env, $debug);
--- a/web/app.php
+++ b/web/app.php
@@ -3,7 +3,7 @@
 use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+$loader = require_once __DIR__.'/../app/autoload.php';
 
 // Enable APC for autoloading to improve performance.
 // You should change the ApcClassLoader first argument to a unique prefix
--- a/web/app.php
+++ b/web/app.php
@@ -1,16 +1,19 @@
 <?php
 
-use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/autoload.php';
+/**
+ * @var Composer\Autoload\ClassLoader $loader
+ */
+$loader = require __DIR__.'/../app/autoload.php';
+include_once __DIR__.'/bootstrap.php.cache';
 
 // Enable APC for autoloading to improve performance.
 // You should change the ApcClassLoader first argument to a unique prefix
 // in order to prevent cache key conflicts with other applications
 // also using APC.
 /*
-$apcLoader = new ApcClassLoader(sha1(__FILE__), $loader);
+$apcLoader = new Symfony\Component\ClassLoader\ApcClassLoader(sha1(__FILE__), $loader);
 $loader->unregister();
 $apcLoader->register(true);
 */
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -18,7 +18,7 @@ if (isset($_SERVER['HTTP_CLIENT_IP'])
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
 }
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+$loader = require_once __DIR__.'/../app/autoload.php';
 Debug::enable();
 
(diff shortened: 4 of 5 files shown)


weaverryan:
This makes a lot of sense to me - make `autoload.php` anything that's always shared between the front controllers, console and tests.

Pierstoval:
:+1: , I always wondered why the bootstrap and the autoload were "merged". In dev I don't use the bootstrap, even if it is longer to load.

Tobion:
It would be safe to merge this in 2.3 actually.

fabpot:
That introduces an infinite loop between the autoloader and the bootstrap file. So, this can only be merged with versions of the bootstrap file that does not have the require on the autoload file. Which means changing the distribution bundle first, making a release, and then change the requirement in composer here.

stof:
and the change in the DistributionBundle also requires this, otherwise the autoloader is not loaded at all anymore.

But I'm -1 on this implementation: this would mean that phpunit would now be loading the bootstrap.php.cache file as it is common to have ``app/autoload.php`` loaded in the Bootstrap (which can break as some classes may have been loaded already by PHPUnit). We would have the same issue with the Behat Symfony2Extension (and there, I'm 100% sure that loading the bootstrap.php.cache file will break because of Symfony classes being loaded earlier in Behat as it uses several components).

If you want to split it, I think the front controller should be responsible for loading the autoloader and then the bootstrap file when wanted. This would solve many issues:

- no infinite loop (when using the current version of the bundle, you just load ``vendor/autoload.php`` twice, but this file is meant to be loadable as many times as you want to get access to the ClassLoader instance), no need for synchronized updates of the bundle and the SE
- easier disabling of the bootstrap file when debugging (http://symfony.com/doc/current/cookbook/debugging.html#disabling-the-bootstrap-file-and-class-caching): you just have to comment the line in the front controller rather than having to change it
- easy way to load the autoloading without the bootstrap file for tests (the same way than today)

Tobion:
> That introduces an infinite loop between the autoloader and the bootstrap file. 

It does not as the files are only included `once`

> and the change in the DistributionBundle also requires this, otherwise the autoloader is not loaded at all anymore.

I don't see what you mean.

> this would mean that phpunit would now be loading the bootstrap.php.cache file

This was already the case before. This is the reason why I included the bootstrap in the autoload.

Pierstoval:
> > and the change in the DistributionBundle also requires this, otherwise the autoloader is not loaded at all anymore.

> I don't see what you mean.

The bootstrap file is generated by the SensioDistributionBundle, so it seems it has to be updated according to this update.

stof:
> It does not as the files are only included once

in this case, it means that the return value is broken. If the file was already loaded and the inclusion is skipped, this would return ``true`` instead of the ClassLoader.

> This was already the case before

no it was not forced. I know many projects using ``app/autoload.php`` as their PHPUnit bootstrap file, precisely to avoid loading the bootstrap.php.cache.

Tobion:
> no it was not forced. I know many projects using app/autoload.php as their PHPUnit bootstrap file, precisely to avoid loading the bootstrap.php.cache.

It still not forced, people can use whatever autoload file they see fit. This is just the default way, just as it was before.

> If you want to split it, I think the front controller should be responsible for loading the autoloader and then the bootstrap file when wanted.

So how do you include the bootstrap file for phpunit then? Or do you suggest to never do it at all?

> in this case, it means that the return value is broken

For whom is the return value broken? I don't see it. The new way has the return value correctly. And people using the old way, also still have the return value.

app/autoload.php
```diff
@@ -10,4 +12,6 @@
 
 AnnotationRegistry::registerLoader(array($loader, 'loadClass'));
 
+include_once __DIR__.'/bootstrap.php.cache';
```
stof: should be require_once to get errors on missing files, not warnings.

stof:
> It still not forced, people can use whatever autoload file they see fit. This is just the default way, just as it was before.

but this would mean that they have to duplicate the autoload.php file, or switch the autoload.php file to the structure I'm suggesting to avoid loading the bootstrap.php.cache file

> So how do you include the bootstrap file for phpunit then? Or do you suggest to never do it at all?

I don't include it. the bootstrap.php.cache file just duplicate some class definitions to perform a single loading for all of them instead of using autoloading. It is not necessary to use it (it is just a perf optimization). and not using it makes debugging easier.

> For whom is the return value broken? I don't see it. The new way has the return value correctly. And people using the old way, also still have the return value.

web/app.php uses the return value of the file.
Actually, it is even more broken today. Something needing the return value should not use the ``_once`` variants of the function, and do it on a file allowing to include it multiple times. Otherwise it is unreliable.

app/autoload.php
Tobion: why force errors? its optional. even if it cannot be included, the app still works. no need to require it

Tobion: maybe we should even surpress the warning

app/console
```diff
@@ -1,25 +1,30 @@
 #!/usr/bin/env php
 <?php
 
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Debug\Debug;
```
Tobion: to be consistent with the other front controllers

app/console
```diff
@@ -1,25 +1,30 @@
...
 // read http://symfony.com/doc/current/book/installation.html#configuration-and-setup for more information
 //umask(0000);
 
 set_time_limit(0);
 
-require_once __DIR__.'/bootstrap.php.cache';
+/**
+ * @var Composer\Autoload\ClassLoader $loader
+ */
+$loader = require __DIR__.'/../app/autoload.php';
```
Tobion: consistency in (app_dev.php its also unsed)

web/app.php
```diff
@@ -1,16 +1,19 @@
 <?php
 
-use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+/**
+ * @var Composer\Autoload\ClassLoader $loader
+ */
+$loader = require __DIR__.'/../app/autoload.php';
```
Tobion: we must use require as @stof pointed out to be able to rely on the return value

web/app.php
```diff
@@ -1,16 +1,19 @@
...
+include_once __DIR__.'/bootstrap.php.cache';
```
Tobion: we use include instead of require, as the app is even working if the file cannot be loaded.

Tobion:
I moved the boostrap into the frontcontrollers and only in debug mode. So it fully fixes #788

web/app.php
```diff
@@ -1,16 +1,19 @@
...
 
 // Enable APC for autoloading to improve performance.
 // You should change the ApcClassLoader first argument to a unique prefix
 // in order to prevent cache key conflicts with other applications
 // also using APC.
 /*
-$apcLoader = new ApcClassLoader(sha1(__FILE__), $loader);
+$apcLoader = new Symfony\Component\ClassLoader\ApcClassLoader(sha1(__FILE__), $loader);
```
Tobion: lets please not discuss this

hacfi:
The new version of the PR looks much better to me!

web/app_dev.php
```diff
@@ -18,7 +18,10 @@
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
 }
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+/**
+ * @var Composer\Autoload\ClassLoader $loader
+ */
+$loader = require __DIR__.'/../app/autoload.php';
```
weaverryan: Once upon a time, fabpot didn't want the front controllers to work differently. In practice, I like what you've done here. But we would need to be pretty much absolutely sure that there could never be (and has never been any reported) differences between these. Can we be that sure?

Tobion: I don't see what difference you mean that could be related to this line?

weaverryan: The difference is including `bootstrap.php.cache` in `app.php` and `autoload.php` in `app_dev.php`. You're right that both ultimately include `app/autoload.php`, but the risk (if it's even real) is that your `bootstrap.php.cache` is somehow out of sync with your source files, so things work in `app_dev.php` (with the true source files), but don't in `app.php` (which uses the compiled versions). Practically speaking, vendor files are only updated with Composer, which includes the script to re-update `bootstrap.php.cache`, so I'm not sure it's a real issue.

weaverryan: But, ezpublish, for example, already does this (loading a different file based on the environment): https://github.com/ezsystems/ezpublish-community/blob/master/web/index.php#L24. If they haven't had issues, I would think we also would be ok.

app/console
```diff
@@ -1,25 +1,30 @@
...
-use Symfony\Component\Debug\Debug;
-
 $input = new ArgvInput();
 $env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
 $debug = getenv('SYMFONY_DEBUG') !== '0' && !$input->hasParameterOption(array('--no-debug', '')) && $env !== 'prod';
 
 if ($debug) {
     Debug::enable();
+} else {
+    include_once __DIR__.'/bootstrap.php.cache';
```
fabpot: I would even not include the bootstrap for the console as it is less important anyway (and as you removed it from phpunit as well, that's not inconsistent either).

fabpot:
Thank you @Tobion.

web/app.php
fabpot: path is wrong, fixed in d06d448

PR #856, opened 2015-09-17
remove configurator routing

Tobion:
As per https://github.com/sensiolabs/SensioDistributionBundle/pull/217

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
-        "sensio/distribution-bundle": "~4.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3",
         "symfony/monolog-bundle": "~2.4",
-        "sensio/distribution-bundle": "~4.0",
+        "sensio/distribution-bundle": "~5.0@dev",
```
Tobion: the the config.php file will be overwritten when lock file gets updated

stof:
:+1: We should ship with the 5.0 version of the bundle without the WebConfigurator in Symfony 2.8

fabpot:
Thank you @Tobion.

PR #857, opened 2015-09-23
git ignore the whole var directory by default

Tobion:


fabpot:
Makes sense indeed.

fabpot:
Thank you @Tobion.

jvasseur:
Using the "!" prefix to re-include a file in an ignored directory doesn't work. This change make the gitkeep files ignored.

Pierstoval:
It works, else nobody would use it, because by default all files are taken in account, so if you are ignoring some, and want specific files not to be ignored, it means they **are** inside an ignored directory.
 
Where have you tested this for it not to be working properly? :confused:

jvasseur:
I tested this by creating a new project with the content of this repo and doing a ```git add .```, and the gitkeep files weren't added to the repo.

The feature work if you do something like this:
```
\var\cache\*
!var\cache\.gitkeep
```
Because the cache directory is not ignored but his content is so you can negate the ignoring for one file.

But if you do something like this
```
\var\cache\
!var\cache\.gitkeep
```
It doesn't work because git doesn't list the content of an ignored directory for performance reasons.

jvasseur:
BTW, this also make the ```SymfonyRequirements.php`` file ignored when it should be kept.

Pierstoval:
Maybe test by replacing `/var/` with `/var/*` in the `.gitignore` file?

jvasseur:
This doesn't solve the problem, it would make the cache, logs and sessions ignored and git will never look inside them to find the gitkeep files.

The only way to factory the individual rules that would work is
```
/var/*/*
/var/bootstrap.php.cache
!var/cache/.gitkeep
!var/logs/.gitkeep
!var/sessions/.gitkeep
```
But that's quite ugly and not a safe way to do it IMO.

Another solution would be to just remove the gitkeep files. Not sure about the session directory but the cache and logs directories are automatically created.

ogizanagi:
See #863, but it's actually ugly...

PR #859, opened 2015-09-24
[3.0] move tests out of src folder

### LINKED ISSUE #855: [3.0] move tests out of src folder
https://github.com/symfony/symfony-standard/pull/850#issuecomment-140676179
> Test classes could be autoloaded only in a development environment by moving them from src/AppBundle/Tests to tests and adding a autoload-dev section in composer.
What do you think about this suggestion?

Pierstoval:
Fixes #855

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,9 @@
     "autoload": {
         "psr-4": { "": "src/" }
     },
+    "autoload-dev": {
+        "psr-4": { "Tests\\": "tests/" }
+    },
     "require": {
         "php": ">=5.5.9",
         "symfony/symfony": "3.0.x-dev",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,7 +9,7 @@
 >
     <testsuites>
         <testsuite name="Project Test Suite">
-            <directory>tests/*</directory>
+            <directory>tests</directory>
         </testsuite>
     </testsuites>
 
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,9 +9,7 @@
 >
     <testsuites>
         <testsuite name="Project Test Suite">
-            <directory>src/*/*Bundle/Tests</directory>
-            <directory>src/*/Bundle/*Bundle/Tests</directory>
-            <directory>src/*Bundle/Tests</directory>
+            <directory>tests/*</directory>
         </testsuite>
     </testsuites>
 
@@ -24,11 +22,8 @@
             <directory>src</directory>
             <exclude>
                 <directory>src/*Bundle/Resources</directory>
-                <directory>src/*Bundle/Tests</directory>
                 <directory>src/*/*Bundle/Resources</directory>
-                <directory>src/*/*Bundle/Tests</directory>
                 <directory>src/*/Bundle/*Bundle/Resources</directory>
-                <directory>src/*/Bundle/*Bundle/Tests</directory>
             </exclude>
         </whitelist>
     </filter>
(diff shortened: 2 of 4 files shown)


phpunit.xml.dist
```diff
@@ -9,9 +9,7 @@
 >
     <testsuites>
         <testsuite name="Project Test Suite">
-            <directory>src/*/*Bundle/Tests</directory>
-            <directory>src/*/Bundle/*Bundle/Tests</directory>
-            <directory>src/*Bundle/Tests</directory>
+            <directory>tests/*</directory>
```
stof: ``tests`` is enough actually

stof:
no need for a ``.gitkeep`` in ``tests`` IMO

Pierstoval:
Fixed in b83f70e, thanks @stof

Pierstoval:
We may also have to change the [sensiolabs/SensioGeneratorBundle](https://github.com/sensiolabs/SensioGeneratorBundle) `generate:bundle` command in order to add the tests directly in this new directory, or the generated default tests would not be automatically added to the test suite

gnugat:
Looks good to me, the choice of `Tests` root namespace allow less constraints than `AppBundle\Test`.
:+1:

tests/Controller/DefaultControllerTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace AppBundle\Tests\Controller;
+namespace Tests\Controller;
```
fabpot: I would have put that into `AppBundle` instead: `Tests\AppBundle\Controller`

Pierstoval: Done in b71ca58, thanks @fabpot

Tobion:
The autoload-dev does not match the test namespace.

Pierstoval:
@Tobion Fixed it in 2856777, if it's what you meant :)

composer.json
```diff
@@ -6,6 +6,9 @@
     "autoload": {
         "psr-4": { "": "src/" }
     },
+    "autoload-dev": {
+        "psr-4": { "Tests\\": "tests/" }
```
xabbuh: Why not use `AppBundle\\Tests\\` here so that we can keep using the `AppBundle\Tests` namespace prefix in the test classes?

stof: Well, if you define multiple bundles, this would make it harder to follow the structure (you would have to change the structure of the testsuite)

Pierstoval: I don't know... Forcing the namespace to `Tests` instead of an empty one sounds better to me because there should not be any other thing than tests.
 
If we wanted to use `AppBundle\Tests` it would need the PSR-4 to be ` { "" : "test/" } ` and we could put almost anything in it.
 
Plus, the bundle structure is merely optional, even if @fabpot advised adding it.
At first I prepared the structure as `tests/Controller/DefaultControllerTest.php` because here we have the whole application tests, and the bundle structure seemed less important to me than the tests we may add in it.

Tobion:
:+1: good to merge IMO

The remaining thing already has an issue on sensiolabs/SensioGeneratorBundle#416

weaverryan:
I share @stof's concern: I recommend the single bundle approach, but I also think adding multiple bundles shouldn't require reworking your structure.

fabpot:
@weaverryan What would you have to change in a multiple-bundle approach?

weaverryan:
@fabpot nevermind - I don't know what I was looking at. `Tests\AppBundle` clearly is laid out perfectly for multi-bundles: :+1:

fabpot:
Thank you @Pierstoval.

PR #861, opened 2015-09-28
[WIP] commented usage of the templating sub-system

fabpot:
For 2.8, I'd like to promote a Twig-only approach. For this to happen:
- ~~We need to get rid of AsseticBundle (#860)~~
- We need to bump minimum versions of our deps (DoctrineBundle and SwitfmailerBundle) as a Twig-only approach only work when using the `@Some/.../....html.twig` notation

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -19,9 +19,8 @@ framework:
     csrf_protection: ~
     validation:      { enable_annotations: true }
     #serializer:      { enable_annotations: true }
-    templating:
-        engines: ['twig']
-        #assets_version: SomeVersionScheme
+    #templating:
+        #engines: ['php', 'twig']
     default_locale:  "%locale%"
     trusted_hosts:   ~
     trusted_proxies: ~


weaverryan:
@fabpot would this deprecate the `AppBundle:Default:index.html.twig` syntax and thus `@App/Default/index.html.twig` would become the recommended way?

fabpot:
@weaverryan It's more than that: without templating, the `AppBundle:Default:index.html.twig` syntax does not work at all, so `@App/Default/index.html.twig` would be the only way to reference templates. That's why we are using only the "new" notation internally (and third-party bundles should do the same by the way).

weaverryan:
@fabpot I thought that might be the case and I would also like to see that. I dislike the `AppBundle:Default:index.html.twig` syntax and the template locator stuff is quite convoluted. I do worry a little bit about people using `@AppBundle` on accident (which iirc resolves to the bundle root instead of Resources/views - or maybe it won't now?), but I have no evidence that this will be a real problem :).

fabpot:
sensiolabs/SensioFrameworkExtraBundle#431 is also a prerequisite as version 3 relies exclusively on the templating component.

Then, reading the Symfony docs, it should be updated as it still recommends using the old notation instead of the native Twig one.

weaverryan:
@fabpot issue opened for the docs about that: symfony/symfony-docs#7003

Pierstoval:
Does this PR have to be closed after SF4 + Flex announcements?

stof:
@weaverryan for Twig templates, `@AppBundle` does not resolve to anything by default

PR #863, opened 2015-10-07
Fix .gitignore for var dir

ogizanagi:
Yeah, it's actually pretty ugly, but AFAIK, it's the only way to keep the whole `var/` directory ignored, except those specific files in subdirectories.
Also, since the recently merged https://github.com/symfony/symfony-standard/pull/857, `SymfonyRequirements.php` is being ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,16 @@
 /app/config/parameters.yml
 /build/
 /phpunit.xml
-/var/
+/var/*
+!/var/cache
+/var/cache/*
 !var/cache/.gitkeep
+!/var/logs
+/var/logs/*
 !var/logs/.gitkeep
+!/var/sessions
+/var/sessions/*
 !var/sessions/.gitkeep
+!var/SymfonyRequirements.php
 /vendor/
 /web/bundles/


Pierstoval:
What about this:

```diff
-/var/*
-!/var/cache
/var/cache/*
!var/cache/.gitkeep
-!/var/logs
/var/logs/*
!var/logs/.gitkeep
-!/var/sessions
/var/sessions/*
!var/sessions/.gitkeep
+/var/*
!var/SymfonyRequirements.php
```

Could it work by moving `/var/*` at the end of the file? I don't know if Git takes the order in account :confused: It could be a bit better-looking..

ogizanagi:
I tried your example to be sure...but it doesn't work and exclude the whole `/var/` dir.

Pierstoval:
It's strange that simply removing the `!/var/sessions` and alikes does not work :confused: 
Indeed it's ugly, but it makes me wonder of how `.gitignore` *really* works...

DavidBadura:
why do you need to keep specific files or folders? these are generated automatically if they are missing. I have some projects for a long time with this config. it work's without problems.

hacfi:
I think it’s good idea to check in the directories to avoid permission issues. Will check the PR later today.

DavidBadura:
the whole var folder should be writable by the web server. it should be no problem to create the missing files and folders.

jvasseur:
@DavidBadura The var/bootstrap.php.cache should probably not be writable by the web server.

DavidBadura:
we persist thumbnails, uploads etc. in this folder which are stored by the web server. instead of insert all folders and files into the gitignore. we should think about to remove files they are not writable by the web server from the var folder. it's just `var/bootstrap.php.cache` right?

Tobion:
Thank you @ogizanagi.

Tobion:
I tested this and it indeed fixes the problem. This appraoch is also descibed in http://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder
It's ugly but the way it has to be to keep the cache/log/session dir as we currently do.

Btw, there would be an alternative: instead of using .gitkeep we use a .gitignore in the cache/log/session dir with

```
*
!.gitignore
```

and the root gitingore would only do `!/var/cache`. But at the end it's the same amount of lines and having it at root in one place is often considered better.

PR #864, opened 2015-10-08
Update .gitignore

hoangbienfgc:


hoangbienfgc:
Need get all file

Pierstoval:
:interrobang:

hoangbienfgc:
I want install symfony 2.7 on my server. Please help me. Thank

javiereguiluz:
@hoangbienfgc to install Symfony 2.7 in your server you just need to install the Symfony Installer and execute this command: `symfony new my_project 2.7`  Everything is explained in http://symfony.com/doc/current/book/installation.html

Pierstoval:
Plus, you should never add to your git repository the files that are ignored in `.gitignore`. If they are ignored, it's because there's a good reason for it.

stof:
-1 for this. It makes no sense to remove these rules from the standard edition.

If you want to commit everything in your project (including cache and logs files in your case, which is probably totally wrong), do it in your project alone, not for all Symfony projects created in the future by anyone. We need to provide a sane ignore list by default.

xabbuh:
:-1: for all the reasons already given

Issue #865, opened 2015-10-09
Wrong in project bundle or wrong phpunit configuration for 2.3


amansilla:
I'm not sure about what is the default bundle for the version 2.3, since actually is `AppBundle`. For version 2.4 return to the old default bundle `Acme/DemoBundle` and from 2.5 again `AppBundle`.

I guess the default bundle name was changed by mistake on version 2.3 otherwise, I think it should be also applied to version 2.4.

In case the default bundle name for version 2.3 is right, the phpunit configuration should be fixed.

Guys, let me know yours thoughts. :smile:

xabbuh:
The change was made at a time when Symfony 2.5 was supported, but 2.4 was not anymore. Thus, changes only got merged up from the `2.3` branch to `2.5` and higher.

amansilla:
Ok thanks for the info @xabbuh. 

In that case my pull request https://github.com/symfony/symfony-standard/pull/866 should be applied since the test of the AppBundle aren't going to be executed, isn't?

PR #866, opened 2015-10-09
Update phpunit configuration

amansilla:
| Q             | A
| ------------- | ---
| Doc fix?      | no
| New docs?     | no
| Applies to    | 2.3
| Fixed tickets | #865

Fix a problem since the bundle without vendor name directly
inside the src directory are not going to be tested.

### DIFF
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -11,6 +11,7 @@
         <testsuite name="Project Test Suite">
             <directory>../src/*/*Bundle/Tests</directory>
             <directory>../src/*/Bundle/*Bundle/Tests</directory>
+            <directory>../src/*Bundle/Tests</directory>
         </testsuite>
     </testsuites>
 


app/phpunit.xml.dist
```diff
@@ -11,6 +11,7 @@
         <testsuite name="Project Test Suite">
             <directory>../src/*/*Bundle/Tests</directory>
             <directory>../src/*/Bundle/*Bundle/Tests</directory>
+            <directory>../src/*Bundle/Tests</directory>
```
xabbuh: We can move this up so that we have the same order here as in the filter below.

amansilla: :+1:

xabbuh:
:+1:

stof:
:+1:

Tobion:
Please change the order of the directories to be the same as in 2.7 to ease merging: https://github.com/symfony/symfony-standard/blob/2.7/app/phpunit.xml.dist

stof:
and please update the code coverage whitelist too (take it from the 2.7 branch too)

amansilla:
@stof the code coverage whitelist is already up to date I guess, isn't?

app/phpunit.xml.dist
xabbuh: Ah sorry, I did not see that we had it in the other order in the `2.7` branch.

amansilla:
Is there still any code changes needed or everything is fine?

Tobion:
Thank you @amansilla.

PR #867, opened 2015-10-11
To do dynamic the database driver too

sinner:
Update config.yml and parameter.yml.dist to do dynamic the database driver too.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -39,7 +39,7 @@ twig:
 # Doctrine Configuration
 doctrine:
     dbal:
-        driver:   pdo_mysql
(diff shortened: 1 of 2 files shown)


app/config/config.yml
```diff
@@ -39,7 +39,7 @@ twig:
 # Doctrine Configuration
 doctrine:
     dbal:
-        driver:   pdo_mysql
+        driver:   "%database_driver%"
```
fabpot: I think this is a bad idea as having two environment use different DB drivers sounds scary.

Tobion:
This was actually removed on purpose before.

sinner:
It's right

PR #869, opened 2015-10-13
Autoload the AppKernel using Composer

### LINKED ISSUE #868: Why is AppKernel not loaded by Composer?
Is there a reason why the `AppKernel` file is not loaded by Composer? I.e. registered like this in `composer.json`:

```json
{
    ...
    "autoload": {
        ...
        "files": [
            "app/AppKernel.php"
        ]
    },
    ...
}
```

Since it is always loaded, letting Composer do the job would make each entry point of the application (console, app.php, app_dev.php, …) a little simpler.

mnapoli:
Fixes #868

Using Composer to load the AppKernel means we don't have to manually include it in all the entry points of the application.

### DIFF
--- a/app/console
+++ b/app/console
@@ -15,7 +15,6 @@ set_time_limit(0);
  * @var Composer\Autoload\ClassLoader $loader
  */
 $loader = require __DIR__.'/../app/autoload.php';
-require_once __DIR__.'/AppKernel.php';
 
 $input = new ArgvInput();
 $env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'dev');
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -24,8 +24,6 @@ if (isset($_SERVER['HTTP_CLIENT_IP'])
 $loader = require __DIR__.'/../app/autoload.php';
 Debug::enable();
 
-require_once __DIR__.'/../app/AppKernel.php';
-
 $kernel = new AppKernel('dev', true);
 $kernel->loadClassCache();
 $request = Request::createFromGlobals();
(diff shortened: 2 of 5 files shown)


Tobion:
:+1:

app/console
```diff
@@ -15,7 +15,6 @@ set_time_limit(0);
  * @var Composer\Autoload\ClassLoader $loader
  */
 $loader = require __DIR__.'/../app/autoload.php';
```
xabbuh: This looks like it can be simplified too.

mnapoli: This is not part of this pull request, additionally I believe the variable is kept so that you can uncomment the code below: https://github.com/symfony/symfony-standard/blob/2.8/web/app.php#L11-L19

Tobion:
ping @symfony/deciders

aitboudad:
@mnapoli I think you missed https://github.com/symfony/symfony-standard/blob/2.8/web/app_dev.php#L27

web/app_dev.php
```diff
@@ -24,8 +24,6 @@
 $loader = require __DIR__.'/../app/autoload.php';
 Debug::enable();
 
-require_once __DIR__.'/../app/AppKernel.php';
```
hacfi: @aitboudad it’s there

mnapoli:
@aitboudad see https://github.com/symfony/symfony-standard/pull/869#discussion_r42759424

web/app_dev.php
aitboudad: :)

aitboudad:
:+1:

xabbuh:
:+1:

Tobion:
Thank you @mnapoli.

xabbuh:
Shouldn't we backport this to older branches too (I don't see a reason why new Symfony 2.3 or 2.7 applications should not benefit from this change).

Tobion:
It's not a bugfix, so IMO it shouldn't be backported.

PR #872, opened 2015-10-20
Add iconv to requirements

nicolas-grekas:
ext-iconv is already built by default into php since years (same as ext-json). Nobody has seen a php without iconv since years :) If we agree on this, then I could remove the iconv shim from https://github.com/symfony/symfony/pull/16240 and we could write code relying on `iconv*()` the same way we do for `json_encode`: without thinking about the function existence.


nicolas-grekas:
Wrong repos, see https://github.com/sensiolabs/SensioDistributionBundle/pull/230

PR #875, opened 2015-10-22
Make sure tests are run with full error reporting

Tobion:
This is also done in all symfony components. It makes sure that tests report all errors and warnings including strict ones.
Often this is not the case otherwise (because it's not part of the php defaults). The other advantage is that it makes tests more equal in PHP 5 and PHP 7. PHP  7 removed strict errors and changed them to warnings or notices. So in PHP 7 those errors might cause failures but they do not cause failures in PHP 5 when strict is ignored.

### DIFF
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -7,6 +7,10 @@
          colors="true"
          bootstrap="bootstrap.php.cache"
 >
+    <php>
+        <ini name="error_reporting" value="-1" />
+    </php>
+
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>../src/*/*Bundle/Tests</directory>


Tobion:
@nicolas-grekas what do you think?

fabpot:
:+1:

Tobion:
I wonder how this interacts with https://github.com/symfony/symfony-standard/blob/2.8/app/autoload.php#L6 that is part of the bootstrapping?

Pierstoval:
@Tobion I think this line should be specific to entry points (`app.php`, `app_dev.php` and `console`) and not the autoloader :confused:

Tobion:
@nicolas-grekas do you agree with this PR? And shouldn't we add symfony/phpunit-bridge to the require-dev section as well?

nicolas-grekas:
LGTM Yes. Adjoint the bridge is a good idea although not strictly required since it looks like that no deprecations are triggered for now

Tobion:
I just tested it with phpunit and the php config is executed before AND after the `bootstrap` file. So the error_reporting is -1 when calling autoload and is reset to -1 after it by phpunit. But it is not set for each test. This is strange behavior by phpunit. But at the end it doesn't really matter to us. There is no difference if error reporting is `-1` or `-1 & ~E_USER_DEPRECATED` if you use the phpunit bridge.

So this is good to be merged.

Tobion:
For the phpunit-bridge I created #884

fabpot:
Thank you @Tobion.

PR #877, opened 2015-10-27
Update checks in SymfonyRequirements for PHP 5.3.9

mbabker:
Updates the checks in the SymfonyRequirements class to match the Symfony 2.7 PHP minimum.


Pierstoval:
This change should be in [SensioDistributionBundle](https://github.com/sensiolabs/SensioDistributionBundle) as the requirements file is generated from the [skeleton file](https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/skeleton/app/SymfonyRequirements.php) in this bundle.

mbabker:
Learned something new today.  Got it.

PR #878, opened 2015-10-28
Same as #852 for 3.0

GrahamCampbell:
I'm not sure if this will even be needed if you update this when doing a 2.8 -> master merge, but it's here anyway if you want it. :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     },
     "require": {
         "php": ">=5.5.9",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
// cc @Tobion

Tobion:
I already did  that when merging.

GrahamCampbell:
Great. I suspected you would. :)

PR #883, opened 2015-11-13
[app] escapes session save path dynamic path for consistency and for preventing any parsing issues.

hhamon:


Tobion:
Thank you @hhamon.

PR #884, opened 2015-11-14
Add phpunit-bridge to handle deprecations

Tobion:
Even though the standard-edition itself doesn't/shouldn't trigger deprecations, the phpunit bridge should still be included by default as developers will have to deal with deprecations sooner or later anyway (and they could even come from outside of symfony).

I targeted the 2.7 branch because that is where the phpunit-bridge was released and where symfony added deprecation warnings.


fabpot:
Thank you @Tobion.

PR #891, opened 2015-11-29
Move the placeholder for assets version in config.yml

kachkaev:
Uncommenting ’#assets_version: SomeVersionScheme’ will cause an exception in Symfony 2.7+ The parameter was moved to framework -> asserts.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -21,7 +21,6 @@ framework:
     #serializer:      { enable_annotations: true }
     templating:
         engines: ['twig']
-        #assets_version: SomeVersionScheme
     default_locale:  "%locale%"
     trusted_hosts:   ~
     trusted_proxies: ~
@@ -32,6 +31,7 @@ framework:
     fragments:       ~
     http_method_override: true
     assets: ~
+        #version: SomeVersionScheme
 
 # Twig Configuration
 twig:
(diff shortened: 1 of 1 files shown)


app/config/config.yml
```diff
@@ -32,6 +31,7 @@ framework:
     fragments:       ~
     http_method_override: true
     assets: ~
+        #version: SomeVersionScheme
```
fabpot: If the user only uncomment this line, it won't work as he also needs to remove the `~` above. Instead, I would just remove the whole comment above (`assets_version`)

kachkaev: This makes sense, but hides the fact that the assets can (and should) be versioned. Not sure what is better / worse.

kachkaev:
Docs need updating too as they refer to the old ``assets_version``:
http://symfony.com/doc/current/reference/configuration/framework.html#ref-framework-assets-version

Perhaps, mentioning [assets version bundle](https://github.com/kachkaev/KachkaevAssetsVersionBundle) there will be useful too.

Tobion:
:+1: 

Status: Reviewed

Tobion:
Should be merged in 2.7 as that is where this config has been deprecated. But this will conflict currently as `assets` config in SE has not been added there yet.

weaverryan:
:+1:

fabpot:
Thank you @kachkaev.

PR #893, opened 2015-11-29
remove security-acl from dependencies

Tobion:
I guess it was just added in https://github.com/symfony/symfony-standard/commit/82658ac4a1c5735d044f5b805761bb33aa5b15a1 to resolve unstable dependecies which should not be required anymore with https://github.com/doctrine/DoctrineCacheBundle/pull/77


fabpot:
Thank you @Tobion.

PR #895, opened 2015-11-30
use semver caret operator and stable symfony 3

Tobion:
@fabpot would be nice to merge this before tagging 3. all those libraries use semantic versioning. so lets use caret operator to spreads best practices.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,19 +12,19 @@
     },
     "require": {
         "php": ">=5.5.9",
-        "symfony/symfony": "3.0.*@dev",
-        "doctrine/orm": "~2.5",
-        "doctrine/doctrine-bundle": "~1.6",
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @Tobion.

JHGitty:
@Tobion Could you explain the different between tildes (~) and carets (^)? I don't really understand the difference in the official composer documentation. https://getcomposer.org/doc/articles/versions.md#caret

stof:
@JHGitty if you use only a major or minor version after the operator, there is no difference. But if you specify a patch version, ``~1.3.2`` will go only up to the next minor version (``<1.4``), i.e. it allows to increment only the last segment (excluding major versions), whereas ``^1.3.2`` will still allow to move up to the next major version (``<2.0``)

PR #897, opened 2015-12-01
Use proper autoloading for the kernel rather than file inclusion

stof:
This allows to register the autoloading for AppCache too, making the diff smaller when wanting to use it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "description": "The \"Symfony Standard Edition\" distribution",
     "autoload": {
         "psr-4": { "": "src/" },
-        "files": [ "app/AppKernel.php" ]
+        "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
     },
     "require": {
         "php": ">=5.3.9",
--- a/web/app.php
+++ b/web/app.php
@@ -18,8 +18,6 @@ $loader->unregister();
 $apcLoader->register(true);
 */
 
-//require_once __DIR__.'/../app/AppCache.php';
-
 $kernel = new AppKernel('prod', false);
 $kernel->loadClassCache();
 //$kernel = new AppCache($kernel);


Pierstoval:
Very :+1:

stof:
sorry to have missed it before the release. I noticed files was used instead of classmap while looking at the KnpUniversity screencast about upgrading a project.

Tobion:
:+1:

xabbuh:
:+1:

weaverryan:
Makes sense to me. @stof is this just a "nice" thing, or are there some real consequences related to autoloading (related to your conversation on symfony/symfony-installer#213)?

stof:
There is a consequence: the AppKernel class is not loaded as soon as you load the composer ClassLoader, but only on first usage, allowing to keep its parent classes in the bootstrap.php.cache file as done previously.

the nice side-effect is also making it simpler to use AppCache as it is now autoloaded

Tobion:
Thank you @stof.

JHGitty:
This fix is not included if I create a new project with the symfony installer tool.
```bash
# Linux, Mac OS X
$ symfony new test

# Windows
c:\> php symfony new test
```

I think a lot of people (for example beginners) will not know this issue.

stof:
@JHGitty this is expected, as this has been merged *after* the 2.8 release, and so will be available in the 2.8.1 archives

xabbuh:
It it isn't too much work it would nonetheless be good to rebuild the archives with the current SE to reduce bad experiences for anyone starting new projects based on 2.8.0 or 3.0.0 in the meantime.

jvasseur:
@xabbuh that would mean recreating the tags which is probably a bad idea.

javiereguiluz:
Besides, even for small changes, recreating the published packages is scary because unexpected things could break.

JHGitty:
That because I hope there will be a new release (2.8.1, 3.0.1, ..) as soon as possible :+1:

Pierstoval:
As I asked on twitter, a new patch version is released approximately every month, so we can expect 2.8.1 to be released in less than 20 days :+1: :wink:

PR #898, opened 2015-12-03
Replace old array() syntax with new []

bocharsky-bw:
We could use a new array syntax `[]` based on the fact that Symfony 3 require PHP `>=5.5.9`

### DIFF
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -10,6 +10,6 @@ error_reporting(error_reporting() & ~E_USER_DEPRECATED);
  */
 $loader = require __DIR__.'/../vendor/autoload.php';
 
-AnnotationRegistry::registerLoader(array($loader, 'loadClass'));
+AnnotationRegistry::registerLoader([$loader, 'loadClass']);
 
 return $loader;
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -12,7 +12,7 @@ use Symfony\Component\Debug\Debug;
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', 'fe80::1', '::1')) || php_sapi_name() === 'cli-server')
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
 ) {
     header('HTTP/1.0 403 Forbidden');
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -7,7 +7,7 @@ class AppKernel extends Kernel
 {
     public function registerBundles()
     {
-        $bundles = array(
+        $bundles = [
             new Symfony\Bundle\FrameworkBundle\FrameworkBundle(),
             new Symfony\Bundle\SecurityBundle\SecurityBundle(),
             new Symfony\Bundle\TwigBundle\TwigBundle(),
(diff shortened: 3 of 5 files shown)


Pierstoval:
I don't know, I like it like it is, I don't like the new syntax (reminds me of Javascript too much, and is a pain to write on Mac rather than the classic syntax).

calinpristavu:
It's true that the old syntax is "prettier", however IMO it's just a matter of habit.
PHPCS also sniffs it and throws a warning.
so I'm +1

bocharsky-bw:
@Pierstoval Why does it pain for you to write new array syntax on Mac?

Almost any application is written with PHP and JS, so to have the same syntax is more readable IMHO.

Pierstoval:
It's a matter of taste, I won't blame anyone for using it, but I personally don't like it because on mac I always confound `{` and `[` (and reversely `}` and `]` ) when I write code...

Tobion:
We don't use the new sytax in symfony/symfony just to ease merging. But we can easily use it in symfony-standard. :+1:

gnugat:
> @calinpristavu 
>
> PHPCS also sniffs it and throws a warning.

But there's no reason to throw a warning: the "old" syntax hasn't been deprecated or anything. The only issue would be mixing both syntaxes.

> @bocharsky-bw 
>
> Almost any application is written with PHP and JS

Symfony can be used as a full stack framework with PHP and javascript mixed, but that's not the only usage. It can also be used to create APIs where you wouldn't have any javascript. It could be used in many ways where you wouldn't have javascript. Also there's no way to know which usage is the most prevalent.

> @Pierstoval
>
> It's a matter of taste

I agree. It doesn't matter which one we use as long as we're consistent everywhere.

bocharsky-bw:
> I agree. It doesn't matter which one we use as long as we're consistent everywhere.

It means that I need to use old syntax in my project if I want to be consistent for now. But why I should do it? The main benefit of new array syntax is that it's shorter than old syntax. So why don't use it if we can.

app/autoload.php
```diff
@@ -5,11 +5,11 @@
 
 error_reporting(error_reporting() & ~E_USER_DEPRECATED);
 
-/**
- * @var ClassLoader $loader
+/*
+ * @var ClassLoader
```
Tobion: these changes should be reverted to actually work in IDEs. fabbot.io is wrong.

web/app_dev.php
```diff
@@ -12,14 +12,14 @@
...
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
 ) {
     header('HTTP/1.0 403 Forbidden');
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
 }
 
-/**
- * @var Composer\Autoload\ClassLoader $loader
+/*
+ * @var Composer\Autoload\ClassLoader
```
Tobion: please revert

app/autoload.php
bocharsky-bw: OK, reverted.

Pierstoval:
> @gnugat It means that I need to use old syntax in my project if I want to be consistent with Symfony SE for now. But why I should do it? The main benefit of new array syntax is that it's shorter than old syntax. So why don't use it if we can.

It's all about standards. Php-fig and PSRs were created in order to simplify writing "beautiful" (rather than "good") PHP applications, and having standards for coding is better to avoid one developer to use tabs instead of spaces, or a developer wanting to put opening brackets in the same line as the declaration, or in the following line.
 
For 20 years (exagerating ? :p ) we wrote `array()`, so everyone know this syntax, this why it's "standard". The short-syntax was implemented in 5.4 and many users don't even know about traits...

xabbuh:
I don't care about this as it dos not really matter whether or not we use the short syntax (people could also change these few occurrences on their own if they feel disturbed by it).

fabpot:
Thank you @bocharsky-bw.

PR #906, opened 2015-12-10
Remove special care for E_USER_DEPRECATED

nicolas-grekas:
We don't need this anymore

### DIFF
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -3,8 +3,6 @@
 use Doctrine\Common\Annotations\AnnotationRegistry;
 use Composer\Autoload\ClassLoader;
 
-error_reporting(error_reporting() & ~E_USER_DEPRECATED);
-
 /**
  * @var ClassLoader $loader
  */


xabbuh:
:+1:

stof:
:+1:

Tobion:
IMO its still better to keep it as some libraries do not use silencing with `@tigger_error` like Elastica.

nicolas-grekas:
We added this in 2.8, elastica runs on 2.7, thus it already works without silencing

Tobion:
The problem is E_USER_DEPRECATED is not excluded by default and thus if you don't use Symfony Debug ErrorHandler, user deprecations are taken action on, e.g. in production mode.
This might be not wanted (since E_DEPRECATED is excluded by default).

nicolas-grekas:
Nobody reported any issue with 2.7, thus I'd say this is theoretical.
But if we keep this on the long run, people *will* ignore things that they cant ignore right now.

Tobion:
> But if we keep this on the long run, people will ignore things that they cant ignore right now.

I don't understand.

Btw, how would it be possible currently to log symfonys deprecations in production mode? I guess one has to use the Debug ErrorHandler somehow in production mode?
It's unfortunate that the silencing makes it impossible to use php internal `log_errors` directive, doesn't it?

nicolas-grekas:
2.8.0 is the only released version where deprecations are silenced, this is new behavior, but bad imho.
Apps running on non-2.8.0 already fixed/handled deprecations, since they have no choice but deal with them (when unsilenced)
Making it impossible to log deprecations in prod with the internal error handler is not an issue at all for me (and not related to this PR btw).

Tobion:
I don't see why you think this is new behavior. You added this silencing in https://github.com/sensiolabs/SensioDistributionBundle/pull/201 and thus is it part even of the newer 2.3 releases of symfony-standard. This is not new to 2.8.0 at all. I just extracted it from bootstrap to autoload.

I'm ok to remove that again. I'm just poiting out there will be a different behavior for deprecations triggered via `@trigger_error` and `trigger_error` in production mode.

stof:
@Tobion non-silenced deprecations would then go to error_log in prod, which will not break your app

nicolas-grekas:
> is it part even of the newer 2.3 releases

Oh really? Then I somehow messed up :-)
Still, to me, this is just accounting for the fact that we silence
deprecations now...

Tobion:
@stof unless people use an error handler in prod that transforms deprecations to exceptions. This to my knowledge exactly the reason why we silenced deprecations in the first place. As workaround for all those error handlers that treated deprecations like any other error.

In the hope that those error handlers are fixed by now, I'm :+1: to merge this.

nicolas-grekas:
On 2.3 isn't it?

Tobion:
This can only be merged in 2.8. The 2.3 I was referring to was about the sensio distrubution bundle bootstrap generation (which is used in 2.3). Nicolas, you need to take a break :) You are getting confused.

nicolas-grekas:
For sure! Christmas time will help :-)

Tobion:
Thank you @nicolas-grekas.

PR #908, opened 2015-12-14
Simplify the log files by ignoring the "event" channel by default

javiereguiluz:
Imagine that you are a newcomer who installs Symfony. You browse the "Welcome Page" and everything seems to work fine:

symfony_welcome_page

But then, you open the `var/logs/dev.log` file and find this monstrosity:

```
[2015-12-14 15:05:08] event.DEBUG: Notified event "console.command" to listener "Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure". [] []
[2015-12-14 15:05:08] event.DEBUG: Notified event "console.command" to listener "Symfony\Bridge\Monolog\Handler\ConsoleHandler::onCommand". [] []
[2015-12-14 15:05:08] event.DEBUG: Notified event "console.command" to listener "Symfony\Bridge\Monolog\Handler\ConsoleHandler::onCommand". [] []
[2015-12-14 15:05:12] request.INFO: Matched route "homepage". {"route_parameters":{"_controller":"AppBundle\\Controller\\DefaultController::indexAction","_route":"homepage"},"request_uri":"http://127.0.0.1:9000/"} []
[2015-12-14 15:05:12] security.INFO: Populated the TokenStorage with an anonymous Token. [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure". [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Component\HttpKernel\EventListener\DumpListener::configure". [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Bundle\FrameworkBundle\EventListener\SessionListener::onKernelRequest". [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Component\HttpKernel\EventListener\FragmentListener::onKernelRequest". [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Component\HttpKernel\EventListener\RouterListener::onKernelRequest". [] []
[2015-12-14 15:05:12] event.DEBUG: Notified event "kernel.request" to listener "Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest". [] []
[...]

PR #911, opened 2015-12-15
ignore event and doctrine logs in console

Tobion:
Follow up on #908 to also ignore event logs in console. And simplify #707 to just ignore doctrine logs as well. #707 was not good because it creates inconsistencies between prod and dev environment. I.e. `-vv` has different result in prod and dev environment because the verbosity_level map is different which is really unexpected if you want to process stderr log in prod environment but they are not there.

So this gets rid these messages when running `app/console -vvv`

```
[2015-12-15 16:40:04] event.DEBUG: Notified event "console.command" to listener "Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure". [] []
[2015-12-15 16:40:04] event.DEBUG: Notified event "console.command" to listener "Symfony\Bridge\Monolog\Handler\ConsoleHandler::onCommand". [] []
```

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -21,18 +21,7 @@ monolog:
         console:
             type:   console
             bubble: false
-            verbosity_levels:
-                VERBOSITY_VERBOSE: INFO
-                VERBOSITY_VERY_VERBOSE: DEBUG
-            channels: ["!doctrine"]
-        console_very_verbose:
-            type:   console
-            bubble: false
-            verbosity_levels:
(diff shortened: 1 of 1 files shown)


javiereguiluz:
:+1: nice! Thanks @Tobion.

fabpot:
LGTM (/cc @nicolas-grekas)

Pierstoval:
:+1: So much code to remove for good :)

fabpot:
Thank you @Tobion.

PR #914, opened 2015-12-18
Use the getRootDir method instead of __DIR__

kmelia:
Hi,

I think it would be better to use the method ``getRootDir()``, what do you think about it?

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -35,12 +35,12 @@ class AppKernel extends Kernel
 
     public function getCacheDir()
     {
-        return dirname(__DIR__).'/var/cache/'.$this->getEnvironment();
+        return dirname($this->getRootDir()).'/var/cache/'.$this->getEnvironment();
     }
 
     public function getLogDir()
     {
-        return dirname(__DIR__).'/var/logs';
+        return dirname($this->getRootDir()).'/var/logs';
     }
 
     public function registerContainerConfiguration(LoaderInterface $loader)


app/AppKernel.php
```diff
@@ -35,12 +35,12 @@ public function getRootDir()
 
     public function getCacheDir()
     {
-        return dirname(__DIR__).'/var/cache/'.$this->getEnvironment();
```
bocharsky-bw: Do we really need to use `dirname()` here?

stof: yeas, to avoid calling ``realpath``. Returning a non-normalized path can prevent Symfony from avoiding absolute paths in the cache files

Tobion:
What is the avantage? It just adds 2 method calls.

kmelia:
If you change the root directory or use a different path, the cache and logs directories will follow the change.

Tobion:
That's not necessarily true. You can place the logs and cache dir whereever you want. They are not bound to the root dir at all. This is why I don't agree with this change.

Pierstoval:
I'll add that it seems to be common that logs are stored in `/var/logs`

kmelia:
> You can place the logs and cache dir whereever you want. They are not bound to the root dir at all.

If you have your own kernel, you will override ``getLogDir`` and ``getCacheDir`` so you do what you want.

> That's not necessarily true.

Maybe you prefer a private method who returns ``dirname(__DIR__)`` used in the 3 public methods? Instead of calling the root directory method. 

It will be more understandable.

PR #917, opened 2015-12-21
use another method to provide more understandable content to getRootDir, ...

kmelia:
… getCacheDir and getLogDir

This is the proposition I made on the last comment of the closed PR #914 

It provide more understandable content to the ``get*Dir()`` methods.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -30,17 +30,22 @@ class AppKernel extends Kernel
 
     public function getRootDir()
     {
-        return __DIR__;
+        return $this->getProjectDir().'/app';
     }
 
     public function getCacheDir()
     {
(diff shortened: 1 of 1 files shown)


Tobion:
It adds 4 methods calls on every request for no reason. So I'm :-1:

Tobion:
Btw, you can do these changes in your own AppKernel easily. This is just a template after all.

kmelia:
> Btw, you can do these changes in your own AppKernel easily. This is just a template after all.

Yes of course, it was just the proposition I could not do in PR #914 ;)

jakzal:
:-1: 

I agree with @Tobion. There's no benefit of doing this while we'd add method calls in a class that's used on every request.

PR #918, opened 2015-12-23
Referencing scripts in composer.json to avoid duplication

1ed:
It's a new feature of composer ([referencing-scripts](https://getcomposer.org/doc/articles/scripts.md#referencing-scripts)) available as of [1.0.0-alpha11](https://github.com/composer/composer/releases/tag/1.0.0-alpha11).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/phpunit-bridge": "^2.7"
     },
     "scripts": {
-        "post-install-cmd": [
+        "symfony-scripts": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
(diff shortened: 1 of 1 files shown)


Pierstoval:
It forces users to update their composer installation, and even if composer itself sends warnings, I don't know if it's a best practice.
But anyway I really like this feature :+1:

xabbuh:
Given that it's included in a tagged release and that release is even more than a month old I am :+1: on this.

WouterJ:
Please note that the Composer update warnings are no shown after 60 days instead of 30, maybe it's better to wait another month?

xabbuh:
Oh did they exceed that value?

1ed:
We already use `config.platform` from the same release as of https://github.com/symfony/symfony-standard/commit/d0e9308e807aae02c332897d610ad5eb49d81c1e.

Pierstoval:
I think we can merge this now :+1:

fabpot:
That's very different. If config.platform is not taken into account, that's no big deal. Here, if script references are not supported, hooks do not work anymore.

fabpot:
Thank you @1ed.

PR #919, opened 2015-12-24
remove references to Symfony2

ip512:
it's weird to have references to sf2 when installing sf3 project

### DIFF
--- a/bin/symfony_requirements
+++ b/bin/symfony_requirements
@@ -7,7 +7,7 @@ $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
 $iniPath = $symfonyRequirements->getPhpIniConfigPath();
 
-echo_title('Symfony2 Requirements Checker');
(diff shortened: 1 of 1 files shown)


Pierstoval:
Duplicate of sooo many other PRs, it's solved in SensioDistributionBundle and will be fixed on the next release

ip512:
I don't see any symfony_requirements file in SensioDistributionBundle but anyway, if you say it's fixed...

Pierstoval:
The file is [here](https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/skeleton/app/SymfonyRequirements.php) and the zip files that one downloads from the website are generated by [this script](https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/bin/build.sh) which is executed server-side by SensioLabs

ip512:
indeed, thanks for the information!

PR #923, opened 2016-01-03
[RFC][DX] Add test front controller

Nek-:
Hello,

I suggest to add this new front controller because if you try to use `php bin/console server:run -e=test` you're website doesn't work (blank page) and the error is not obvious at all ! See:

```
[2016-01-03 02:15:25] php.DEBUG: fsockopen(): unable to connect to 127.0.0.1:8000 (Connection refused) {"type":2,"file":"/.../vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Command/ServerCommand.php","line":59,"level":28928} []
```

I lost some time searching for the reason, and as symfony core supports test env by default, the standard edition should support it too.

I put the same "security" inside the app_test file (that is in the same way bypassable), for similar reasons.

Hope you'll agree.

### DIFF
--- a/web/app_test.php
+++ b/web/app_test.php
@@ -1,0 +1,34 @@
+
+<?php
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Debug\Debug;
+
+
+// If you don't want to setup permissions the proper way, just uncomment the following PHP line
+// read http://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup
+// for more information
+//umask(0000);
+
+// This check prevents access to debug front controllers that are deployed by accident to production servers.
+// Feel free to remove this, extend it, or make something more sophisticated.
+if (isset($_SERVER['HTTP_CLIENT_IP'])
+    || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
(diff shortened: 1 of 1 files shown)


Pierstoval:
> I suggest to add this new front controller because if you try to use `php bin/console server:run -e=test` you're website doesn't work (blank page) and the error is not obvious at all !

It is normal, the `test` environment should be used only when instanciating a kernel within PHPUnit tests.
 
`app_test` would be a web entry, which is not at all the point of using the `test` environment.

jakzal:
We have rejected such requests before, see #734 for explanation.

Nek-:
> It is normal, the test environment should be used only when instantiating a kernel within PHPUnit tests.

Did you try functional tests ? (behat)

@jakzal I understand the fact that you don't want the server to work by default. But if you think so, this file become a problem: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Resources/config/router_test.php

Because it allows to start the integrated server (that is useful in the case to test things). There is no error... Until you try to get a page and it's blank. This behavior should not happens. I mean, **everything is green, no error, even in verbose mode**, but it fails.

And thinking that the `app_test.php` file is missing is all but obvious.

Please consider taking a decision on a side or the other, but the situation is definitly not normal for now.

PR #928, opened 2016-01-13
Update `.gitignore` with rules for `bin` directory

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

This allows to keep track of `bin/console` while leaves the rest of the directory's
content out of SCM, which is often generated by composer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 /app/config/parameters.yml
 /build/
+/bin/*
+!/bin/console
+!/bin/symfony_requirements
 /phpunit.xml
 /var/*
 !/var/cache


bocharsky-bw:
I think this changes only useful after switching to the new symfony 3 directory structure, but instead you better should manually remove all vendor's binary files from the bin dir and use them from vendor/bin.

Tobion:
as @bocharsky-bw also said, SE 3.0 does not put vendor bins in `bin` anymore but in the default `vendor/bin`. `bin` is only for your application specific executables. So closing.

phansys:
Excellent! I missed that point absolutely.
Thanks guys.

PR #930, opened 2016-01-18
update composer.lock file

xabbuh:
Someone noticed on IRC that the current lock file for 3.1.x-dev was not really usable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,8 @@
             "php": "5.5.9"
         }
     },
+    "prefer-stable": true,
+    "minimum-stability": "dev",
     "extra": {
         "symfony-app-dir": "app",
         "symfony-bin-dir": "bin",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -49,6 +49,8 @@
             "php": "5.5.9"
         }
     },
+    "prefer-stable": true,
+    "minimum-stability": "dev",
```
xabbuh: This is needed for the APCu polyfill.

fabpot:
Thank you @xabbuh.

composer.json
Pierstoval: Isn't this really dangerous? I mean, isn't there a way to release a proper release for the APCu polyfill, even a beta/rc for the minimum stability to rise up and not be `dev`?

xabbuh: Imo that's okay for the moment as Symfony 3.1 itself isn't stable yet. But yeah having a release for the APCu polyfill will be better.

Pierstoval: This must not be forgotten when `3.1` will reach stabilization/release phase :confused:

fabpot: That's how we did that in the past for previous versions.

Pierstoval: Ok no problem for that then :)

PR #932, opened 2016-01-20
SensioGeneratorBundle not found for composer install --no-dev

marius-rizac:
When running `composer install --no-dev` or `composer update --no-dev` this error is shown 
PHP Fatal error:  Class "Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle" not found in ...app/AppKernel.php on line 25
because `sensio/generator-bundle` is a development dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,10 +20,10 @@
         "symfony/monolog-bundle": "^2.8",
         "sensio/distribution-bundle": "^5.0",
         "sensio/framework-extra-bundle": "^3.0.2",
+        "sensio/generator-bundle": "^3.0",
         "incenteev/composer-parameter-handler": "^2.0"
     },
     "require-dev": {
-        "sensio/generator-bundle": "^3.0",
         "symfony/phpunit-bridge": "^2.7"
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


xabbuh:
The SensioGeneratorBundle is only loaded in the application kernel if you run it in the `dev` environment. And if you access the application in the `dev` environment, you should have installed the dev dependencies too.

marius-rizac:
@xabbuh please run `composer install --no-dev` on a new cloned directory. And when you run composer install, which environment is used ?

bocharsky-bw:
@marius-rizac By default used `dev` environment. Composer does not share environment with Symfony. You can change it with `export SYMFONY_ENV=prod`. Check http://symfony.com/doc/current/cookbook/deployment/tools.html#c-install-update-your-vendors for more info.

Pierstoval:
+1 for @xabbuh and @bocharsky-bw, and I'll add that if you install with `--no-dev` you might be using a production machine so you should export the env var directly to your machine's environment.

marius-rizac:
@bocharsky-bw ok, got it. thanks.

PR #937, opened 2016-02-04
Properly lock down app_dev.php

phihag:
Currently, web/app_dev.php in the symfony-standard package limits requests to those coming from CLI and unproxied localhost.

This prevents an attacker to execute this file in a production environment.

IPv4 localhost requests can come from 127.0.0.0/8, so the current check is actually a little bit too restrictive, but safe.

IPv6 localhost requests come from ::1/128.

The address fe80::1, however, is a link-local IPv6 address like any other. Anyone in the local network can take it and communicate with it. Therefore, it does not belong here.

### DIFF
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -12,7 +12,7 @@ use Symfony\Component\Debug\Debug;
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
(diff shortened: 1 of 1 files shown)


Pierstoval:
Isn't `fe80::1` the equivalent of `127.0.0.1` in IPV6?

phihag:
@Pierstoval No, see the text in the commit and pull request description. The closest equivalent to `127.0.0.1` (which is just one out of many [loopback addresses](https://en.wikipedia.org/wiki/Localhost) in IPv4) would be `::1`, [the only loopback address in IPv6](https://tools.ietf.org/html/rfc4291#section-2.5.3). `fe80::1` is a [link-local](https://tools.ietf.org/html/rfc4291#section-2.5.6) - [that is to say interface-specific] - address with no special properties. If you think Wikipedia, the IETF, and I a are all mistaken about `fe80::1`, try it yourself - nobody will answer `ping fe80::1%eth0` until you configure the address.

Tobion:
:+1:  Should be merged in 2.3

xabbuh:
:+1:

fabpot:
Thank you @phihag.

PR #940, opened 2016-02-06
Update symfony_requirements

JarJak:
fix symfony version

### DIFF
--- a/bin/symfony_requirements
+++ b/bin/symfony_requirements
@@ -7,7 +7,7 @@ $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
 $iniPath = $symfonyRequirements->getPhpIniConfigPath();
 
-echo_title('Symfony2 Requirements Checker');
+echo_title('Symfony3 Requirements Checker');
 
 echo '> PHP is using the following php.ini file:'.PHP_EOL;
 if ($iniPath) {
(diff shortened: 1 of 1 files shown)


hacfi:
@JarJak Good catch! I think they prefer just `Symfony` though. Let’s wait for feedback though.

bin/symfony_requirements
```diff
@@ -43,9 +43,9 @@ foreach ($symfonyRequirements->getRecommendations() as $req) {
 }
 
 if ($checkPassed) {
-    echo_block('success', 'OK', 'Your system is ready to run Symfony2 projects');
```
hacfi: While you’re at it: you could update `/** @var $req Requirement */` to the more common variant `/** @var Requirement $req */` if you don’t mind

JarJak: both are acceptable by IDE, i don't think your version is more common

Tobion:
Already fixed in https://github.com/sensiolabs/SensioDistributionBundle/pull/247

bin/symfony_requirements
hacfi: @JarJak According to the specs at http://phpdoc.org/docs/latest/references/phpdoc/tags/var.html that is the recommended version and is consistent with `@param ClassName $var`.

JarJak:
but it's not fixed in this repo

Tobion:
it will be updated on the next vendor update automatically.

PR #946, opened 2016-02-18
Raising required PHP version for Symfony3 to 5.5.9

jzawadzki:
Hi,

not sure why but required PHP version in `var/SymfonyRequirements.php` is still 5.3.3.
This value is important as is used by e.g. symfony installer for checking sysstem requirements and it's important from DX perspective (See issue https://github.com/symfony/symfony-installer/issues/239)

This pull request raise the minimum PHP version to 5.5.9 according to http://symfony.com/doc/current/reference/requirements.html

JZ

### DIFF
--- a/var/SymfonyRequirements.php
+++ b/var/SymfonyRequirements.php
@@ -376,7 +376,7 @@ class RequirementCollection implements IteratorAggregate
  */
 class SymfonyRequirements extends RequirementCollection
 {
-    const REQUIRED_PHP_VERSION = '5.3.3';
+    const REQUIRED_PHP_VERSION = '5.5.9';
 
     /**
      * Constructor that initializes the requirements.


jzawadzki:
However, not sure if this should be fixed here or https://github.com/sensiolabs/SensioDistributionBundle/blob/3.0/Resources/skeleton/app/SymfonyRequirements.php#L379 or both

bocharsky-bw:
Yes, this should be fixed in SensioDistributionBundle too. Otherwise the `var/SymfonyRequirements.php` should be overwritten on next composer install/update

Tobion:
This must be fixed in SensioDistributionBundle. But as the probably do not want to maintain different branches for symfony 2.x and 3.x, it needs to be configurable. I would propose that the appropriate version needs to be used depending on the directory structure that is determined in the SensioDistributionBundle which then copies the files.

Tobion:
Fixed in sensiolabs/SensioDistributionBundle#280

PR #947, opened 2016-02-24
Updating symfony/phpunit-bridge version constraint

Jean85:
I recently installed Symfony 3.0.2, done a `composer update` and discovered (using vinkla/climb) that symfony/phpunit-bridge couldn't be updated to the latest major version. This should make the upgrade immediately executable on a fresh install.

I didn't update the lock because v3 increases the minimum PHP version to 5.5.9, so this remains backward compatible with older PHP versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "sensio/generator-bundle": "^3.0",
-        "symfony/phpunit-bridge": "^2.7|^3.0"
+        "symfony/phpunit-bridge": "^3.0"
     },
     "scripts": {
         "post-install-cmd": [
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "sensio/generator-bundle": "^3.0",
-        "symfony/phpunit-bridge": "^2.7"
+        "symfony/phpunit-bridge": "^2.7|^3.0"
     },
     "scripts": {
         "post-install-cmd": [


composer.json
```diff
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "sensio/generator-bundle": "^3.0",
-        "symfony/phpunit-bridge": "^2.7"
+        "symfony/phpunit-bridge": "^2.7|^3.0"
```
bocharsky-bw: What about to use `"symfony/phpunit-bridge": "^3.0"` for Symfony `3.0`?

Pierstoval: +1 here we are on the master branch so no need to comply with 2.x branch.

Jean85: Right! Revising PR now, `^2.7` is needed when backporting..

Jean85:
Since we are on master and with SF3.x, I updated this PR to request phpunit-bridge `^3.0`. `^2.7` is needed when backporting to older branches

Tobion:
:+1: Could be merged in 3.0

xabbuh:
:+1:

fabpot:
Thank you @Jean85.

Jean85:
You're welcome @fabpot !
But... Shouldn't the `composer.lock` be updated now? And with which PHP version? The lowest? I will happily PR that.

stof:
@Jean85 we generally update run ``composer update`` before each release (after the symfony/symfony release)

Jean85:
Ok great!

PR #948, opened 2016-02-24
merged the two php tags in phpunit config file into one

tifabien:


bocharsky-bw:
:+1:

xabbuh:
:+1:

Pierstoval:
:+1:

WouterJ:
:+1:

Tobion:
This should be changed in 2.3: https://github.com/symfony/symfony-standard/blob/2.3/app/phpunit.xml.dist#L24

fabpot:
As the fix is different in 2.3, I've made the change in ea679b0 and merge it up to all other branches. Thanks.

PR #951, opened 2016-03-02
Require ocramius/proxy-manager ^1.0 - 2.0 requires PHP 7

hacfi:
ocramius/proxy-manager 2.0 requires PHP 7 so by default this project should require ^1.0 to not bump the dependency to PHP 7. If someone installs it on a PHP 7 machine and then wants to deploy it on PHP 5.5/5.6 it will fail. Only noticed this because I got an error on our build server after locally updating to 2.0 and pushing the changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "symfony/monolog-bundle": "^2.8",
         "sensio/distribution-bundle": "^5.0",
         "sensio/framework-extra-bundle": "^3.0.2",
-        "incenteev/composer-parameter-handler": "^2.0"
+        "incenteev/composer-parameter-handler": "^2.0",
+        "ocramius/proxy-manager": "^1.0"
     },
     "require-dev": {
         "sensio/generator-bundle": "^3.0",


bocharsky-bw:
Even when you require `^1.0` and then locally updating to `2.0` and pushing the changes - you still will be failed, don't you? Composer reads dependencies from `composer.lock` locked file, so it seems pointless. If you don't want it failed - use the right [platform](https://github.com/symfony/symfony-standard/blob/master/composer.json#L49) in version `composer.json` when updating dependencies locally.

Pierstoval:
As said @bocharsky-bw, use the right `platform` version, this is precisely the reason why the `platform` parameter was introduced.

stof:
Closing this Pr as the change is indeed wrong. If you run different PHP versions locally and on your server, you should indeed use the composer feature dedicated at supporting such use case.

There is no reason to depend on ``ocramius/proxy-manager`` in symfony-standard btw.

hacfi:
Didn’t know about the `platform` config value. This change doesn’t make sense then indeed.

PR #953, opened 2016-03-11
Reduce touch icon to 32 colours and optimize

rjmunro:
Cuts it to about 1/8th it's former size. Visually hardly any difference.


javiereguiluz:
The circle of the new icon looks a bit jagged. How does it look when displayed in a browser tab? Thanks!

Pierstoval:
Actually there *is* a slight difference but I don't know if it's a problem.

capture d ecran 2016-03-11 a 12 46 24

javiereguiluz:
@Pierstoval thanks, but I wanted to see the icon in the browser tab at the mini-size the user will see. Here it is the comparison (top, current icon; bottom, new proposal):

favicon_comparison

I'd say that the new proposal is noticeable worse at small sizes :disappointed:

rjmunro:
I've made a 32 color version - it's about 80% bigger, but still about 1/5 of the size of the original image.

fabpot:
Thank you @rjmunro.

PR #954, opened 2016-03-11
Do security:check after installing or upgrading

rjmunro:
Compared to [adding the `roave/security-advisories` meta-package](https://github.com/symfony/symfony-standard/pull/760):
- It's output is much nicer
- It runs on any install, not just update, so you are more likely to
  get the warning earlier
- It's just a warning - it doesn't block you carrying on with
  upgrading / installing other things

I had a situation where a not-very-critical security error meant that
roave blocked me upgrading other packages, some of which had more
important issues to think about.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
-            "bin/console security:check -q || app/console security:check --ansi || true"
+            "CHECK_OUTPUT=`bin/console security:check --ansi` && echo 'No known vulnerabilities found in vendors' || echo \"$CHECK_OUTPUT\""
         ],
         "post-update-cmd": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -33,7 +33,8 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget"
+            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
+            "app/console security:check -q || app/console security:check --ansi || true"
```
Pierstoval: Should be `bin/console` for Symfony3

composer.json
```diff
@@ -41,7 +42,8 @@
...
```
Pierstoval: Should be `bin/console` for Symfony3

Pierstoval:
I know that the `-q` option is used to show the result only if there are errors, but running the command twice is cumbersome, running it once and for all should be enough. It also reassures on the fact that the command **is** executed and that there is no security flaw in the app, which is important to know (else, the user won't know the command is executed unless he know what's inside the composer file and how it works).

rjmunro:
@Pierstoval The problem is that the default output of the security:check command is 15 lines long and very colourful for success (no vulnerabilities). This is fine if you are running the security:check command because you want to check the security of your site, but added on to the output of a composer call, it's a bit verbose. For failure, however, I really want the full, multi line, multi colour warning.

As I've been writing this, I've thought of some improvements I can make. I'll update this PR shortly.

Pierstoval:
Then you could use something different that stores the output to a file and echoes it only if the security fails:

```bash
$ $(bin/console security:check > var/logs/check.log 2>&1) || cat var/logs/check.log
```

What do you think?

Pierstoval:
Hmm... I'm thinking about something else: will the `bin/console` work on windows?

rjmunro:
@Pierstoval In the version I have just posted, rather than save to a file, I save to a bash variable. None of this is likely to work on windows, however. I think the best approach will be to write some PHP that does the same thing as the console command, rather than try to use the console command itself. I guess that would fit better in the https://github.com/sensiolabs/security-checker project than here, but then this project could be updated to call it.

Pierstoval:
Nice for the variable :+1: 

For the command itself, I think a new option `--silent-success` or similar may be enough, but I don't know if it's gonna be an agreement in the team :smile:

PR #957, opened 2016-03-17
[app] updated PHPUnit configuration according to the current supported PHPUnit version.

hhamon:
Is it valid that we "support" PHPUnit 4.1 or shall we update to the latest supported version?

### DIFF
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<!-- http://phpunit.de/manual/4.1/en/appendixes.configuration.html -->
+<!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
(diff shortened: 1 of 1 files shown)


fabpot:
:+1:

stof:
:+1: 

@hhamon IMO, we should ask people to run at least 4.8

hhamon:
@stof where should we advertize for it? README file?

stof:
hmm, given this is the standard edition, i.e. the bootstrap for new projects, not sure we need to advocate it in the readme (which gets removed anyway during the project creation). Advocating to use a maintained version is PHPUnit's job after all.

We may add a note in the symfony-doc chapter about testing, but I'm not sure it is worth it.

xabbuh:
We could update the version number in the docs. Somewhere we tell people to use 4.2 or higher.

Tobion:
Thank you @hhamon.

PR #959, opened 2016-03-20
[WIP] Support for controller autoregistration and autowiring

dunglas:
* Integrates [DunglasActionBundle](https://github.com/dunglas/DunglasActionBundle)
* Automatically register "actions" as services and enable autowiring

I've updated the Symfony demo application to show the benefits of this new system: symfony/symfony-demo#315

(edited) Features:

* In most cases, this is not necessary anymore to declare services (they are automatically constructed by the autowiring system - DX, RAD)
* Dependencies are explicit (DX, testability)
* The container is not injected in controllers anymore
* It's possible (but not mandatory) to create framework agnostic controllers (especially when using the PSR-7 Bridge)
* It's possible and recommended (but not mandatory) to use invokable actions (it's not what I've done in my PR on the demo to explicit the upgrade path)
* It's 100% BC with the traditional controller system, 0 app will be broken, and people preferring the old way can still use it
* It's 100% compatible with `@Route` and `@Template` annotations

See symfony/symfony#18193 and symfony/symfony#18193 for the history and the rationale behind this feature.

TODO:

* [ ] Merge symfony/symfony#18193
* [ ] Merge symfony/symfony#18193
* [ ] Merge symfony/symfony#17608
* [x] Fix dunglas/DunglasActionBundle#18

/cc @Ener-Getick

### DIFF
--- a/bin/symfony_requirements
+++ b/bin/symfony_requirements
@@ -7,7 +7,7 @@ $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
 $iniPath = $symfonyRequirements->getPhpIniConfigPath();
 
-echo_title('Symfony2 Requirements Checker');
+echo_title('Symfony Requirements Checker');
 
 echo '> PHP is using the following php.ini file:'.PHP_EOL;
 if ($iniPath) {
@@ -43,9 +43,9 @@ foreach ($symfonyRequirements->getRecommendations() as $req) {
 }
 
 if ($checkPassed) {
-    echo_block('success', 'OK', 'Your system is ready to run Symfony2 projects');
+    echo_block('success', 'OK', 'Your system is ready to run Symfony projects');
 } else {
-    echo_block('error', 'ERROR', 'Your system is not ready to run Symfony2 projects');
+    echo_block('error', 'ERROR', 'Your system is not ready to run Symfony projects');
 
     echo_title('Fix the following mandatory requirements', 'red');
 
--- a/src/AppBundle/Action/DefaultAction.php
+++ b/src/AppBundle/Action/DefaultAction.php
@@ -3,13 +3,20 @@
 namespace AppBundle\Action;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Bundle\FrameworkBundle\Controller\ControllerTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 
-class DefaultAction extends Controller
+class DefaultAction
 {
     use ControllerTrait;
+
+    private $container;
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
 
     /**
      * @Route("/", name="homepage")
@@ -18,7 +25,7 @@ class DefaultAction extends Controller
     {
         // replace this example code with whatever you need
         return $this->render('default/index.html.twig', [
-            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..'),
+            'base_dir' => realpath($this->container->getParameter('kernel.root_dir').'/..'),
         ]);
     }
 }
--- a/src/AppBundle/Action/DefaultAction.php
+++ b/src/AppBundle/Action/DefaultAction.php
@@ -1,17 +1,20 @@
 <?php
 
-namespace AppBundle\Controller;
+namespace AppBundle\Action;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Bundle\FrameworkBundle\Controller\ControllerTrait;
 use Symfony\Component\HttpFoundation\Request;
 
-class DefaultController extends Controller
+class DefaultAction extends Controller
 {
+    use ControllerTrait;
+
     /**
      * @Route("/", name="homepage")
      */
-    public function indexAction(Request $request)
+    public function __invoke(Request $request)
     {
         // replace this example code with whatever you need
         return $this->render('default/index.html.twig', [
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -15,6 +15,7 @@ class AppKernel extends Kernel
             new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
             new Doctrine\Bundle\DoctrineBundle\DoctrineBundle(),
             new Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle(),
+            new Dunglas\ActionBundle\DunglasActionBundle(),
             new AppBundle\AppBundle(),
         ];
 
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -18,6 +18,7 @@ framework:
     form:            ~
     csrf_protection: ~
     validation:      { enable_annotations: true }
+    serializer:      { enable_annotations: true }
     #serializer:      { enable_annotations: true }
     templating:
         engines: ['twig']
@@ -66,3 +67,11 @@ swiftmailer:
     username:  "%mailer_user%"
     password:  "%mailer_password%"
     spool:     { type: memory }
+
+# This config will be removed when https://github.com/dunglas/DunglasActionBundle/issues/18 will be fixed
+dunglas_action:
+    autodiscover:
+        enabled:   true
+        directories:
+            action:  [ Action ]
+            command: []
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -67,11 +67,3 @@ swiftmailer:
     username:  "%mailer_user%"
     password:  "%mailer_password%"
     spool:     { type: memory }
-
-# This config will be removed when https://github.com/dunglas/DunglasActionBundle/issues/18 will be fixed
-dunglas_action:
-    autodiscover:
-        enabled:   true
-        directories:
-            action:  [ Action ]
-            command: []
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -19,7 +19,6 @@ framework:
     csrf_protection: ~
     validation:      { enable_annotations: true }
     serializer:      { enable_annotations: true }
-    #serializer:      { enable_annotations: true }
     templating:
         engines: ['twig']
     default_locale:  "%locale%"
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -1,3 +1,3 @@
 app:
-    resource: "@AppBundle/Controller/"
-    type:     annotation
+    resource: "@AppBundle/Action/"
+    type:     action-annotation
--- a/composer.json
+++ b/composer.json
@@ -10,9 +10,15 @@
     "autoload-dev": {
         "psr-4": { "Tests\\": "tests/" }
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/dunglas/symfony"
+        }
+    ],
     "require": {
         "php": ">=5.5.9",
-        "symfony/symfony": "3.1.*@dev",
+        "symfony/symfony": "dev-action as 3.1.0",
         "doctrine/orm": "^2.5",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/doctrine-cache-bundle": "^1.2",
@@ -20,7 +26,8 @@
         "symfony/monolog-bundle": "^2.8",
         "sensio/distribution-bundle": "^5.0",
         "sensio/framework-extra-bundle": "^3.0.2",
-        "incenteev/composer-parameter-handler": "^2.0"
+        "incenteev/composer-parameter-handler": "^2.0",
+        "dunglas/action-bundle": "dev-master"
     },
     "require-dev": {
         "sensio/generator-bundle": "^3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -779,12 +779,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/dunglas/DunglasActionBundle.git",
-                "reference": "713cca53ac05076e645eaddd75d68d6a6e88db4c"
+                "reference": "db0039f330216d7cd3186c0359b8b2903a8347d8"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/dunglas/DunglasActionBundle/zipball/713cca53ac05076e645eaddd75d68d6a6e88db4c",
-                "reference": "713cca53ac05076e645eaddd75d68d6a6e88db4c",
+                "url": "https://api.github.com/repos/dunglas/DunglasActionBundle/zipball/db0039f330216d7cd3186c0359b8b2903a8347d8",
+                "reference": "db0039f330216d7cd3186c0359b8b2903a8347d8",
                 "shasum": ""
             },
             "require": {
@@ -832,7 +832,7 @@
                 "routing",
                 "symfony"
             ],
-            "time": "2016-03-20 20:40:25"
+            "time": "2016-03-20 21:53:21"
         },
         {
             "name": "incenteev/composer-parameter-handler",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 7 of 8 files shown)


javiereguiluz:
I would agree to make Symfony "more friendly" to DunglasActionBundle ... but the following change is a big :-1: from me:

controller_action

The change may seem "small", but it's a huge change in the way Symfony worked until today. It would require us to make immense changes, such as the documentation and the training materials.

Ener-Getick:
@javiereguiluz the old controller system still works so is it really a problem ?
To satisfy everyone maybe we could include both possibilities in the standard edition: a controller and an action class? WDYT ?

dunglas:
Both changes are optional but: it doesn't prevent controllers wrote "the traditional way" way to work.

You can keep using the `Controller` directory (but controllers will not be registered as services) and using `__invoke` (even when controllers are registered as services) is not mandatory (but should be considered a good practice IMO). It's 100% BC.

As the traditional way of writing controllers is still working well, doc can be updated progressively like it has been done for the new directory structure.

bin/symfony_requirements
```diff
@@ -7,7 +7,7 @@ $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
 $iniPath = $symfonyRequirements->getPhpIniConfigPath();
 
-echo_title('Symfony2 Requirements Checker');
+echo_title('Symfony Requirements Checker');
```
theofidry: should not be part of this PR?

src/AppBundle/Action/DefaultAction.php
```diff
@@ -1,17 +1,20 @@
...
-namespace AppBundle\Controller;
+namespace AppBundle\Action;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Bundle\FrameworkBundle\Controller\ControllerTrait;
 use Symfony\Component\HttpFoundation\Request;
 
-class DefaultController extends Controller
+class DefaultAction extends Controller
```
theofidry: do we still need to extend `Controller`?

dunglas: Good catch, it don't.

bin/symfony_requirements
Ener-Getick: It was automatically updated when running ``composer update``

dunglas: Not sure but as this file is always overwritten I think it doesn't hurt.

src/AppBundle/Action/DefaultAction.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class DefaultAction
+{
+    use ControllerTrait;
+
+    private $container;
+
+    public function __construct(ContainerInterface $container)
```
dunglas: As discussed during the hackday, it would be better to inject the parameter directly but it requires to add some lines in `services.yml`. IMO it's very rare that a controller should use a parameter (parameters must usually be injected in standard services). Maybe the best for here is to remove the use of this parameter?

dunglas:
I've updated the Symfony demo application to show advantage of this new system: symfony/symfony-demo#315

* In most cases, this is not necessary anymore to declare services (they are automatically constructed by the autowiring system - DX, RAD)
* Dependencies are explicit (DX, testability)
* The container is not injected in controllers anymore
* It's possible (but not mandatory) to create framework agnostic controllers (especially when using the PSR-7 Bridge)
* It's possible and recommended (but not mandatory) to use invokable actions (it's not what I've done in my PR on the demo to explicit the upgrade path)
* It's 100% BC with the traditional controller system, 0 app will be broken, and people preferring the old way can still use it
* It's 100% compatible with `@Route` and `@Template` annotations

What do you think @javiereguiluz @WouterJ?

WouterJ:
Let's say this first: I like the idea behind the action style Symfony way. I think it's a nice way to make decoupling from frameworks and moving to some nicer DDD apps easier (framework agnostic controller will never be possible though imo, but that's another topic).

However, I'm not sure if it should be the default or recommended way. Besides the advantages, it also has some disadvantages. One I can think of now: I have to create one class and one file per controller action. Reducing the amount of files one has to create before having a functional working app was one of the main goals of the best practice guide (which is about the only reason `@Route` annotations are recommended).

I think while action-style means *easier* decoupling, using normal controllers is still the RAD-est, easiest way to write your app and I think Symfony Demo and Standard should use it instead. Also, decoupling is needed in about 10% of the apps, for 90% it will only make developers life more difficult. I also believe that controllers shouldn't be unit tested, as controller shouldn't contain logic so there isn't much to test in that unit.

(however, I explicitly voted :confused: instead of :-1: because I'm not very sure about my opinion yet).

theofidry:
>Reducing the amount of files one has to create before having a functional working app was one of the main goals of the best practice guide

I'm having trouble to understand here: isn't the best practice guide supposed to give guidelines on best practices? Reducing the amount of files looks like DX for me, it's not about best practices, whereas splitting a controller, at least, promotes SOLID principles (how many principles are we breaking with a single controller?!).

I can perfectly understand the reluctance of making the "Action way" the default in Symfony (although I would love to), but it should at least be promoted as a best practice IMO.

dunglas:
> One I can think of now: I have to create one class and one file per controller action. 

No, you can use both action-style controllers and have several action methods per class. Take a look at the PR I've opened on the demo: symfony/symfony-demo#315

In term of RAD, not having to declare services in the `services.yml` file is a big win isn't it?

I'm fully open to discuss the directory structure (the bundle can work both with `Controller` or `Action` by default, and both with `__invoke` and multi-method controllers). I propose the current structure because it (IMO) recommend a clean way to creates controllers/action. But what I really want to get merged is *autowired controllers as a services** by default.

WouterJ:
> In term of RAD, not having to declare services in the services.yml file is a big win isn't it?

The first 17 chapters of the Symfony Book are able to create controllers, forms, entities, tests, ... without even introducing the term "service". There is no need to declare services in Symfony using traditional controllers.

dunglas:
@WouterJ you must introduce a service every time you create business logic. And any small real app has business logic. Services are at the core of the Symfony-way of building apps (and it's basically the main change between Symfony 1 and 2).

A lot of Symfony apps have large controllers with ton of actions and all the business logic in controllers. I hope this proposal will encourage Symfony to keep their controllers as small as possible and to introduce services more frequently (we cannot force people to write good code, but we can encourage them to do it). And btw this PR is vaguely inspired by the ADR pattern but has nothing to do directly with DDD.

javiereguiluz:
According to the [Symfony Core Team rules](http://symfony.com/doc/current/contributing/code/core_team.html), :-1: votes must be justified by "technical and objective reasons". Here are my reasons to vote :-1::

  * It deprecates most of the symfony.com documentation and blog posts written about Symfony until today. This will cause confusion in lots of users. It also deprecates most of the existing training materials.
  * The default controller/action showed to developers imply that you must create 1 file/class per action, which increases "perceived complexity" a lot. (However, you can define several actions per file).
  * I've always heard that PHP's magic methods (like `__invoke()`) are slow from a performance point of view (I don't have stats to back this affirmation! Could someone shed light on this?)
  * For newcomers, the new way to work feels "unnatural". Believe or not, most developers have never used a trait and consider "__invoke()" a PHP "rarity".
  * It depends on a bundle which hasn't been battle-tested: [DunglasActionBundle](https://github.com/dunglas/DunglasActionBundle) has 120 stars on GitHub and just 92 installs [according to Packagist](https://packagist.org/packages/dunglas/action-bundle). I'm sure the code of this bundle is good enough because Kévin is a great developer and knows Symfony well ... but integrating this into Symfony will set a worrying precedent.

---

And now a subjective reason:

  * If we adopt such a big change, it must be at least 10 times better than the current situation: the developer must be 10 times more productive or the application 10 times more performant, etc.
  * But this change just proposes to use "traits" instead of "extends" and "__invoke()" instead of "indexAction()". Where's the benefit in productivity or performance or DX? In the past you made `$this->get('slugger')` and now you must do `use ...\Slugger` and create a constructor to get that service. You don't reduce the number of lines to write or the number of concepts to learn.
  * The main arguments in favor of this change are: 1) you don't have to declare services and 2) controllers are turned into services automatically. However:
    * In real applications you probably need to create those services after all because they are used in other services, in console commands, etc. not only on controllers.
    * Most people don't define their controllers as services. I can't see the benefit of it.

---

We could make 3 things related to this:

1. Make Symfony capable of using DunglasActionBundle
2. Integrate DunglasActionBundle in Symfony itself
3. Integrate DunglasActionBundle and make it the default way to use Symfony

This PR goes for 3)  My vote is for 1).

Pierstoval:
I'm not a pro in terms of patterns and all these things, I just comment as a regular Symfony user, and I cannot agree more to what @javiereguiluz said:
Making Symfony capable of using Actions system is clearly enough to allow more flexibility and fit to other patterns.
 
But with my 3-years experience in Symfony 2, I can say that I have *never* declared controllers as services, and *almost always* extended Symfony's controller in my own controllers (I can remember only one or two cases, used only for maximal performances).
 
Refactoring the whole documentation to fit a new design pattern does not sound DX at all to me, because the whole web knows about Symfony's Kernel>Controller>Response system to be a standard and of most common use.
 
Plus, regarding the recent changes in Symfony's releases schedules, I don't see any benefit to integrate such complex default behavior and make it BC in the next minor releases. The new major releases will come soon enough to avoid having "dirty BC" in most versions, and even if it's personal, I'm afraid that this new design pattern standard would *replace* Symfony's one, and I don't want to see classic Controllers disappear in a Symfony 5 release for example (ok, it's a bit paranoid, sorry about that).
 
My "fears" may or may not be reasonable, this is just my comment.

But for all the reasons @javiereguiluz invoked (no ADR pun intended), I'm also :-1: for this :confused:

src/AppBundle/Action/DefaultAction.php
```diff
@@ -0,0 +1,31 @@
...
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\ControllerTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class DefaultAction
+{
+    use ControllerTrait;
+
+    private $container;
```
vincentchalamon:
```php
/**
 * @var ContainerInterface
 */
private $container;
```

src/AppBundle/Action/DefaultAction.php
vincentchalamon:
```php
/**
 * @param ContainerInterface $container
 */
public function __construct(ContainerInterface $container)
```

src/AppBundle/Action/DefaultAction.php
dunglas: Not useful, IDE can guess it with the constructor typehint.

bocharsky-bw: @vincentchalamon :+1: , PHP doc above a property allows do not scan the whole file to find what type is it. The question here is not that IDE can do it for you but improve code readability for other developers. BTW, I suppose not everyone IDE can guess it.

dunglas:
@javiereguiluz OK got most of your points, I reply below and I've proposed an alternative with emphasis on what I mean is the important part for now: #960

> It deprecates most of the symfony.com documentation and blog posts written about Symfony until today. This will cause confusion in lots of users. It also deprecates most of the existing training materials.

It's a very important point. But all examples in doc and training material will still work.
IMO the legacy must not prevent us of moving forward. We need BC, we need an upgrade path (and we have them here), but we need to always improve Symfony.

> The default controller/action showed to developers imply that you must create 1 file/class per action, which increases "perceived complexity" a lot. (However, you can define several actions per file).

I don't agree with you on that point. One of the most common problem in Symfony apps is developers putting all the business logic in controllers. On many large projects, controllers are far too big "god objects" and are a nightmare to maintain.
My proposal promote 3 best practices:

* extract business logic into services
* keep your controllers as small as possible (easier to maintain and to understand for newcomers in the project)
* choose carefully your dependency and don't create god objects

But I can understand that some people like you think it's a too big step for now, it's why I opened #960, but IMO it's better to make both "old" controllers and "new" actions cohabit and promote the new one.

> I've always heard that PHP's magic methods (like __invoke()) are slow from a performance point of view (I don't have stats to back this affirmation! Could someone shed light on this?)

You got a point, it should be benchmarked. I think it's slower but it's negligible for the number of actions called in a request. Btw, renaming `__invoke` to `invoke` or `action` will still work but I find that less elegant.

> For newcomers, the new way to work feels "unnatural". Believe or not, most developers have never used a trait and consider "__invoke()" a PHP "rarity".

Sure, but it's also the mission of Symfony to educate developers to new PHP features and best practices. And the old system still work.

> It depends on a bundle which hasn't been battle-tested: DunglasActionBundle has 120 stars on GitHub and just 92 installs according to Packagist. I'm sure the code of this bundle is good enough because Kévin is a great developer and knows Symfony well ... but integrating this into Symfony will set a worrying precedent.

You're 100% right. It's exactly why I want this merged before the feature freeze: to battle-test it. 120 stars in only 2 months is not so bad IMO (it's a sign of interest for this kind of approach by the community). [IncenteevParameterHandler](https://github.com/Incenteev/ParameterHandler) has only 260 stars and has been available for years.

But I'm open to any solution to ease the maintenance: add the SF core teams to the maintainers, move it under the symfony namespace, merge it in core...

> If we adopt such a big change, it must be at least 10 times better than the current situation: the developer must be 10 times more productive or the application 10 times more performant, etc.

It's not a so big change. Actually, doing that with Symfony worked for years and the bundle is very small. As the old system will still work, even if it's 10% better IMO we should move.

> But this change just proposes to use "traits" instead of "extends" and "__invoke()" instead of "indexAction()". Where's the benefit in productivity or performance or DX? In the past you made $this->get('slugger') and now you must do use ...\Slugger and create a constructor to get that service. You don't reduce the number of lines to write or the number of concepts to learn.

Reducing the number of lines is not always the way to go. But here you actually reduce it (no entry in `services.yml`. It also reduces the number of concept to learn at first (no config file to know with it's own directory structure...). It becomes similar to how Laravel works but with cleaner concepts and extensibility behind.

I've used my bundle on some projects since 2 months and automatic dependency building is very pleasant and time-saver.

Using composition instead of inheritance in such case is also considered universally a best practice.

Easing the migration to framework agnostic controllers is also something that matter in term of code quality and marketing.

> In real applications you probably need to create those services after all because they are used in other services, in console commands, etc. not only on controllers.

The bundle supports commands too thanks to @Ener-Getick (see my updated PR on the demo), other services have their their dependency autowired too. It's the main point about this PR. As controllers and  commands are the entrypoint of the service graph, in 70 to 80% of cases (maybe more), you don't need to declare your services explictly anymore!

With my proposal, controllers are indeed services but it's 100% transparent for the end user. The intention is not to promote controllers as services, but to use them to unlock new features and improve code quality.

@Pierstoval 
> Plus, regarding the recent changes in Symfony's releases schedules, I don't see any benefit to integrate such complex default behavior and make it BC in the next minor releases. The new major releases will come soon enough to avoid having "dirty BC" in most versions, and even if it's personal, I'm afraid that this new design pattern standard would replace Symfony's one, and I don't want to see classic Controllers disappear in a Symfony 5 release for example (ok, it's a bit paranoid, sorry about that).

It's not dirty at all. And I'll be :-1: to remove the traditional controller system in any future Symfony release. It can be useful in a lot of contexts.
Technically both my bundle and the "old" system share the same Symfony internals. I've used the traditional controller system for years and I can't find any argument to remove it. The maintenance will be as complex with or without it.
IMO their is no reason to be afraid of this new proposal replacing the traditional way. Both can cohabit without any problem.

src/AppBundle/Action/DefaultAction.php
dunglas: I've no opinion about that but it's not what we do in symfony/symfony and it's not compliant with our current CS. Let's discuss this in another issue.

src/AppBundle/Action/DefaultAction.php
dunglas: Same as above, not valid according to Symfony CS (and IMO useless).

weaverryan:
Thanks for your work @dunglas. I definitely dislike this - at least as an initially step. But I *am* interested in your #960. Whatever we do, it will need to be *a* way to do things, but not *the* recommended way to do things initially. We have to give ourselves some time using this stuff to see how it feels before we could roll it out as any type of a recommended practice. Personally, I'm very interested, but not convinced of the whole idea - just because we haven't had enough people coding with it long enough.

dunglas:
Closing in favor of #960

PR #964, opened 2016-04-12
dont hint the usage of the ApcuClassLoader anymore

backbone87:
the ApcuClassLoader isnt needed for PHP 5.6+ / PHP 7+ anymore, since
Composer can now dump a OPCache optimized static classmap, that has
similar performance characteristics

also the ApcuClassLoader could cause problems with memory fragmentation
for large apps

see https://github.com/composer/composer/pull/5174

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -1,17 +1,8 @@
 <?php
 
-use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
 $loader = require_once __DIR__.'/../app/bootstrap.php.cache';
-
-// Use APC for autoloading to improve performance.
-// Change 'sf2' to a unique prefix in order to prevent cache key conflicts
-// with other applications also using APC.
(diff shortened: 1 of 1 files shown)


Tobion:
:+1:

Pierstoval:
Still needed for php < 5.6 actually, shouldn't this PR be made over 2.8 instead of 2.3 ?

Seldaek:
IMO if you care about this level of performance you are on PHP5.6 or even 7 by now.. Upgrading PHP is the biggest win, not nitpicking with autoloaders :) 

Also this file is only used by new projects, older projects stuck on 5.4/5.5 will keep their file with the APC instructions, and new projects are more likely to be setup on new servers with up to date PHP. So all in all I'd say :+1: for just removing it.

xabbuh:
:+1:

stof:
:+1:

Tobion:
Thank you @backbone87.

PR #971, opened 2016-05-09
solve code-duplication into composer.json

eugene-matvejev:
Hi Guys,

thanks for your contribution,

I tried find way how I can avoid code duplication what we got in "scripts" space, i found a solution
I think it is worth to evaluate

please don't judge hard

Best Wishes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,15 +27,7 @@
         "symfony/phpunit-bridge": "^3.0"
     },
     "scripts": {
-        "post-install-cmd": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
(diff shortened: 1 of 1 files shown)


stof:
this is wrong. We don't want to run them after each dump of the autoload:

- clearing the cache or installing assets after ``composer dump-autoload`` does not mark sense
- these actions must still run when running ``composer install --no-autoloader``

Removing code duplication is possible by using a custom script to hold them and referencing it in both post-install and post-update: https://getcomposer.org/doc/articles/scripts.md#referencing-scripts

Pierstoval:
I agree with @stof, as scripts referencing is available since [1.0.0-alpha9](https://github.com/composer/composer/releases/tag/1.0.0-alpha9) in November 2015, I think it's time to introduce them in Symfony composer config

fabpot:
:-1: for changing anything here. A bit of duplicate does not hurt, and we still have the possibility to make those sections diverge if needed. Having an external script makes things more complex for no benefits. Let's close this one.

eugene-matvejev:
thanks for feedback 👍

eugene-matvejev:
@stof , @Pierstoval  https://github.com/eugene-matvejev/symfony-standard/pull/1/files

NB! run composer self-update to get at least v1.0.3

that is better solution? (tested, works)

PR #972, opened 2016-05-13
Merge $_SERVER to $_ENV in dev front controller

marfillaster:
php-cli built-in webserver router does not pass mutated $_SERVER to controller

Fixes symfony/symfony#17077


marfillaster:
closing this. change is only necessary when frontend controller is explicitly called in the request, e.g. /app_dev.php. cli server router automatically rewrites all requests

PR #974, opened 2016-05-13
Default system cache is APCu

nicolas-grekas:


xabbuh:
👍

fabpot:
Thank you @nicolas-grekas.

JHGitty:
Why is this only merged to `3.1` and not to `master` ?

Pierstoval:
I think that's because it adds a new dependency?

JHGitty:
But why should new dependencies not be added to master? Shouldn't it be added to new versions like 3.2?

stof:
@JHGitty this is the other way actually. Older branches are merged into newer branches on a regular basis. So if it is merged into 3.1, it will also be available in 3.2 soon.
If it had been merged into master directly, 3.1 would never contain this change.

JHGitty:
Okay, great. Thanks! :+1:

PR #976, opened 2016-05-22
Removing the symfony-tests-dir composer extra config - it's not used

weaverryan:
This was originally proposed for sensiolabs/SensioGeneratorBundle#416 (in #862), but I'm proposing to solve that in a simpler way: sensiolabs/SensioGeneratorBundle#494

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,6 @@
         "symfony-bin-dir": "bin",
         "symfony-var-dir": "var",
         "symfony-web-dir": "web",
-        "symfony-tests-dir": "tests",
         "symfony-assets-install": "relative",
         "incenteev-parameters": {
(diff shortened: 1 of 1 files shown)


fabpot:
Why not use this configuration setting in #494?

weaverryan:
@fabpot afaik, there is no way to read composer.json extra values at runtime (other than parsing the file manually). Plus, are we sure we want the extra `composer.json` key just to help an edge feature in the generator? Unless there are some other places where we might use that value in the future?

Ener-Getick:
Is it worth removing it ? I can think of some use cases in third party library.

fabpot:
Let's keep it, it does not hurt.

PR #977, opened 2016-05-22
Console should not accept an empty --env option

voronkovich:
Now, if you try to run `bin/console --env`, it throws an exception:

    [InvalidArgumentException]
    The file "/path/to/app/app/config/config_.yml" does not exist.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -17,7 +17,7 @@ set_time_limit(0);
 $loader = require __DIR__.'/../app/autoload.php';
 
 $input = new ArgvInput();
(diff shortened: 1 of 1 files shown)


voronkovich:
Also I think it would be great to add a checking to the Kernel constructor. Now the Kernel accepts `''` as a value for the environment. See https://github.com/symfony/http-kernel/blob/master/Kernel.php#L78

jvasseur:
Passing an empty env option should probably fail instead of silently defaulting to dev.

Pierstoval:
I'm :-1: for this change, this should be resolved in the kernel itself, but I think it can introduce BC breaks...

PR #979, opened 2016-06-01
Require bootstrap instead of autoload

Koc:


Pierstoval:
I don't think autoload would have been thrown out of bootstrap in 7311e70 to be reintroduced later :confused:

Koc:
But the first lines of generated bootstrap is:

```php
namespace {
    error_reporting(error_reporting() & ~E_USER_DEPRECATED);
    $loader = require_once __DIR__.'/./autoload.php';
}
```

so it includes autoload.

//cc @Tobion

Tobion:
https://github.com/sensiolabs/SensioDistributionBundle/blob/master/UPGRADE.md#upgrade-from-40-to-50

Koc:
Thank you @Tobion

PR #986, opened 2016-07-21
Filter only private IP allowed

IndraGunawan:
fixing to allow access from private IP to app_dev.php, including when using docker for Mac that use virtualbox as docker-machine. the machine IP like 192.168.99.100 but is still private IP.

but using this condition other people in LAN also can access the app_dev.php


indragunawan:
but using this condition other people in LAN also can access the app_dev.php

Pierstoval:
Not everyone use docker containers or private IPs, so this should depend on the platform and not be part of the standard edition

RicardoJ:
Just like @Pierstoval says, imo, this is not needed as you can change this after a clean or current installation of the standard edition.

indragunawan:
thanks @Pierstoval and @RicardoJ for your attention :) 
i will close this PR.

PR #990, opened 2016-08-30
Change inline var PHPDoc syntax

teohhanhui:
See https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#722-var

(refer to the [examples](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-21))

Related previous PR: #571 (but there was no standard for inline `@var` at the time)


hacfi:
I like it :)

fabpot:
Thank you @teohhanhui.

PR #991, opened 2016-09-01
Display a directory separator adapted to the OS Symfony is running on

Alsciende:
On Windows, the current Symfony installer displays 

> Your application is now ready. You can start working on it at:

and a path followed by `/`, even though the path itself (returned by `realpath`) is separated by `\`. It is inelegant et potentially confusing.

This PR tries to fix that.

### DIFF
--- a/src/AppBundle/Controller/DefaultController.php
+++ b/src/AppBundle/Controller/DefaultController.php
@@ -15,7 +15,7 @@ class DefaultController extends Controller
     {
         // replace this example code with whatever you need
         return $this->render('default/index.html.twig', [
-            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..'),
+            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..').DIRECTORY_SEPARATOR,
         ]);
     }
 }
--- a/src/AppBundle/Controller/DefaultController.php
+++ b/src/AppBundle/Controller/DefaultController.php
@@ -14,8 +14,8 @@ class DefaultController extends Controller
     public function indexAction(Request $request)
     {
         // replace this example code with whatever you need
-        return $this->render('default/index.html.twig', array(
+        return $this->render('default/index.html.twig', [
             'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..').DIRECTORY_SEPARATOR,
-        ));
+        ]);
     }
 }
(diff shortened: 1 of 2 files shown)


src/AppBundle/Controller/DefaultController.php
```diff
@@ -14,8 +14,8 @@ class DefaultController extends Controller
     public function indexAction(Request $request)
     {
         // replace this example code with whatever you need
-        return $this->render('default/index.html.twig', [
-            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..'),
-        ]);
+        return $this->render('default/index.html.twig', array(
+            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..').DIRECTORY_SEPARATOR,
```
bocharsky-bw: But why you missing directory separator in `'/..'`?

src/AppBundle/Controller/DefaultController.php
```diff
@@ -14,8 +14,8 @@ class DefaultController extends Controller
     public function indexAction(Request $request)
     {
         // replace this example code with whatever you need
-        return $this->render('default/index.html.twig', [
-            'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..'),
-        ]);
+        return $this->render('default/index.html.twig', array(
```
bocharsky-bw: We replace old array syntax with a new one in https://github.com/symfony/symfony-standard/pull/898

src/AppBundle/Controller/DefaultController.php
Alsciende: This one is parsed by `realpath`, which returns only `\` on Windows.

src/AppBundle/Controller/DefaultController.php
Alsciende: The "Coding Standard" check from fabbot complains when a PR is submitted with the new syntax. Should I re-submit with the new syntax anyway?

bocharsky-bw: I think we have to keep the new syntax anyway.

jakzal:
:+1:

Alsciende:
Is this PR ignored because of the failing check?

bocharsky-bw:
@Alsciende No, merging just takes some time. And because other developers could be able to give their opinion (props and cons) about these changes as well.

fabpot:
Thank you @Alsciende.

PR #993, opened 2016-09-20
Add note about permissions setup in app.php

bocharsky-bw:
We already have this for `app_dev.php`

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -2,6 +2,11 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
+// If you don't want to setup permissions the proper way, just uncomment the following PHP line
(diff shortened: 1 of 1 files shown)


jakzal:
Don't you think that permissions should be set up the "proper way" on prod?

bocharsky-bw:
Of course, if it possible. There's a link in that comment which refers to the docs where pointed some info about setup permissions in the proper way.

jakzal:
Anyway, I find it messy to have this comment in there. It could be it's justified to have it in the dev controller to make things easier for newcomers. I wouldn't add it to the prod one though.

fabpot:
I agree with @jakzal, it does not make sense to "promote" a bad practice in the production controller.

PR #994, opened 2016-10-02
Finetuned default index

ro0NL:
Any thoughts?

Before:
image

image

After:
image

image

### DIFF
--- a/app/Resources/views/default/index.html.twig
+++ b/app/Resources/views/default/index.html.twig
@@ -2,11 +2,11 @@
 
 {% block body %}
     <div id="wrapper">
-        <div id="container">
-            <div id="welcome">
-                <h1><span>Welcome to</span> Symfony {{ constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION') }}</h1>
-            </div>
+        <div id="welcome">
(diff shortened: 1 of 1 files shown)


ro0NL:
@javiereguiluz wdyt?

app/Resources/views/default/index.html.twig
```diff
@@ -50,22 +50,25 @@
 <style>
     body { background: #F5F5F5; font: 18px/1.5 sans-serif; }
     h1, h2 { line-height: 1.2; margin: 0 0 .5em; }
-    h1 { font-size: 36px; }
+    h1 { font-size: 26px; }
```
javiereguiluz: This value can only be `24` or `28` (we use the traditional typographic scale: 16, 18, 21, 24, 28, 32, 36, 48, 64, 72, 96, etc.)

javiereguiluz:
To be completely honest, I don't like the two main changes proposed in this PR: removing the Welcome message from the main content section and fading some texts in a light gray 😢

fabpot:
@ro0NL We worked a lot on those designs. So, except if there are things that need to be fixed, I don't see the need to change them.

ro0NL:
Your call :) i was just playing around with contrast. Imo it looks better, but this matter of taste. Im not a designer ;-)

Issue #992, opened 2016-09-19
Add a functional Docker environment


dkarlovi:
After having fixed symfony/symfony#19681, add a sample docker-compose.yml file with an example how to set everything up and get up&running really fast.

You could use `nginx:alpine` with `php:7.0-fpm-alpine` to make it really small.

jakzal:
:-1:

I think it's out of scope of this repo. Such configuration is too specific to your project and we have rejected similar requests before (for vagrant at least).

dkarlovi:
@jakzal Vagrant and Docker do not have a 1:1 relationship. Where Vagrant is a development tool, Docker is also a production-ready environment which means you could use exactly the same configuration for deployment.

You could argue "too specific for your project" for Swift mailer too for example, your project doesn't need to send mails, yet you still ship mailing by default.

Pierstoval:
@jakzal:
> 👎
> 
> I think it's out of scope of this repo. Such configuration is too specific to your project and we have rejected similar requests before (for vagrant at least).

Yet there is still a Heroku deployment process in the distribution bundle:

https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Composer/ScriptHandler.php#L96-L117

Why not something similar for Docker?

jakzal:
> Vagrant and Docker do not have a 1:1 relationship. Where Vagrant is a development tool, Docker is also a production-ready environment which means you could use exactly the same configuration for deployment.

I'd say there's no relationship between the two. What I mean is these are tools configured per project and we cannot provide a generic configuration for standard edition to be used by everyone.

dkarlovi:
> I'd say there's no relationship between the two.

Then I don't see the reason to mention declining to include Vagrant config here. :)

> What I mean is these are tools configured per project and we cannot provide a generic configuration for standard edition to be used by everyone.

**Everything** is configured per-project. SE is (the way I see it) **a starting point** which should get you up&running ASAP, but you remove everything you don't use (like the mentioned mailer integration, for example).

HeahDude:
I agree with @jakzal here, maintain docker files in this repo is not a good idea.

We share docker files between devs in our own repo to get started quickly depending on php versions, servers and db, so should you.

Federkun:
> how to set everything up and get up&running really fast.

Well, if you're interested, I made this: https://github.com/Federkun/docker-skeleton-php/wiki/Create-a-new-Symfony-application
It's just a starting point.

fabpot:
[...]

PR #1012, opened 2016-10-08
Adding docker-compose support

rawkode:


bocharsky-bw:
Probably it's more appropriable for Symfony Demo app?

bocharsky-bw:
It took ~1.5s to generate welcome page 😥 Is there a way to speed up it?

screen shot 2016-10-08 at 21 04 09

rawkode:
I'll assume you're on OSX. Running Docker for Mac currently uses osxfs, which has known performance issues atm

bocharsky-bw:
Yes, I'm on Mac. Ah, I hoped that "native" OS X Docker solution would improve performance but seems it didn't happen :(

gmsantos:
@bocharsky-bw try this http://docker-sync.io/, a colleague told me that works well

bocharsky-bw:
Hm, it's something new and mystery to me, but wow, looks like it rocks! It's a bit confusing me, but I found a working solution: https://github.com/timglabisch/symfony_docker_sync and it super fast - only *65ms*! I definitely want to dig deeper. Thank you @gmsantos a lot!!

jakzal:
:-1: for similar reasons given in https://github.com/symfony/symfony-standard/issues/992#issuecomment-247942783 and https://github.com/symfony/symfony/issues/18377#issuecomment-204329546 

Instead, we could add a cookbook to the docs with an example configuration (or add it to the demo app as @bocharsky-bw suggests).

bocharsky-bw:
I think it'd be nice to have some official good example in docs for Docker: a related issue on docs here https://github.com/symfony/symfony-docs/issues/4055 .

PR #1013, opened 2016-10-12
Use PHP's native session handler in dev environment

gerryvdm:
Using the native file session handler with a save path inside the project directory is causing a lot of problems, particularly for Vagrant environments with shared/NFS folders.

I suggest defaulting to PHP's native session handler at least for the development environment.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -5,6 +5,8 @@ framework:
     router:
         resource: "%kernel.root_dir%/config/routing_dev.yml"
         strict_requirements: true
+    session:
+        handler_id:  ~
     profiler: { only_exceptions: false }
 
 web_profiler:


gerryvdm:
See https://github.com/symfony/symfony/issues/17845

Tobion:
👎 to have such a difference for dev and prod env

gerryvdm:
@Tobion as opposed to not working for a considerable share of dev environments?

Pierstoval:
What stats are you using to say that Vagrant setup is a `considerable share of dev environments`?

Even if it's important for vagrant users, a simple cookbook is fairly enough to handle this, just write it in [symfony/symfony-docs](https://github.com/symfony/symfony-docs) if it's not already proposed and make the change.

It's like saying that "docker is very common these days". Yes it is, but Symfony newcomers almost never use Docker nor Vagrant.

Symfony standard edition is here to "bootstrap" an app. If you want to bootstrap it with vagrant, it means you have a "higher" level of set up than common developers.

jakzal:
I agree with @Tobion. It's more something for a cookbook as @Pierstoval suggested. 

It's not even the only way to solve the problem. The existing docs already [suggest moving out the cache and log directories](http://symfony.com/doc/current/setup/homestead.html). We could add a note about sessions.

PR #1023, opened 2016-11-09
Fix invalid links in README.md

yceruto:
https://symfony.com/doc/3.0/book/installation.html

page-not-found

The changes were made based on recent documentation structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,11 @@ released under the MIT or BSD license.
 
 Enjoy!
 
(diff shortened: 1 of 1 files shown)


stof:
Such changes should be done for the 2.7 branch instead (and we will then adapt the version when merging branches together)

stof:
Actually, #1022 has the right fix and was there first, so I'm closing this PR.

PR #1028, opened 2016-12-02
Order uses imports

Lctrs:


fabpot:
Thanks for your PR, but we don't order use statements on old code to avoid making it more difficult branch merges.

PR #1030, opened 2016-12-03
Don't load class cache and bootstrap file on php 7

derrabus:
As discussed in symfony/symfony#20668, the class cache features should be deprecated. Since the class cache and the bootstrap file are still useful on php 5, we load them only when running on php 5.

A future php-7-only release of Symfony Standard should remove those blocks completely.

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -4,10 +4,14 @@ use Symfony\Component\HttpFoundation\Request;
 
 /** @var \Composer\Autoload\ClassLoader $loader */
 $loader = require __DIR__.'/../app/autoload.php';
-include_once __DIR__.'/../var/bootstrap.php.cache';
+if (PHP_VERSION_ID < 70000) {
+    include_once __DIR__.'/../var/bootstrap.php.cache';
+}
 
 $kernel = new AppKernel('prod', false);
-$kernel->loadClassCache();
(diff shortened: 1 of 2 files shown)


derrabus:
ping @nicolas-grekas

derrabus:
The fabbot failure is not related to my changes.

iltar:
When removing this from the repo, this line shouldn't be forgotten: https://github.com/symfony/symfony-standard/blob/1f23b3b1e2fb8c34a7ee1205bbcebbfbf75640ba/composer.json#L33

nicolas-grekas:
:+1:

xabbuh:
👍

jakzal:
:+1:

fabpot:
Thank you @derrabus.

vyshkant:
@iltar 

> When removing this from the repo, this line shouldn't be forgotten:
https://github.com/symfony/symfony-standard/blob/1f23b3b1e2fb8c34a7ee1205bbcebbfbf75640ba/composer.json#L33, 

Could you please explain, what exactly did you mean? Should this line be removed or kept?

linaori:
@vyshkant that line should be removed if you're on php 7.0+, because it builds the bootstrap that you no longer require.

PR #1031, opened 2016-12-06
Set environtment var for app_dev.php

nietzcheson:
E.g.:

```
### ~/.homestead/Homestead.yml
 variables: 
      - key: SYMFONY_DEV
      value: 1
```

### DIFF
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -10,10 +10,10 @@ use Symfony\Component\Debug\Debug;
 
 // This check prevents access to debug front controllers that are deployed by accident to production servers.
 // Feel free to remove this, extend it, or make something more sophisticated.
-if (isset($_SERVER['HTTP_CLIENT_IP'])
+if (!getenv("SYMFONY_DEV") && (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
     || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', '::1']) || php_sapi_name() === 'cli-server')
-) {
+)) {
     header('HTTP/1.0 403 Forbidden');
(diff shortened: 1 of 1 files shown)


Pierstoval:
I don't agree. Homestead is not the "default for standard", and if you're using env vars, there's a lot of chance that you would use `SYMFONY_DEBUG=0|1` and `SYMFONY_ENV=prod|dev` , but it's up to the dev to choose between different env vars.
Plus, with env vars, technically you would not need `app_dev` anymore because everything can be handled with these env vars.

bocharsky-bw:
I agree with @Pierstoval , but anyway, is there somewhere a good example in Symfony docs about how to use a Symfony app during the development along with Docker or other virtualization tools? Because by default, I can't get access to the`app_dev.php` inside Docker container without any changes. It'd be nice to have a good example how to handle it at least in docs if any exists.

Pierstoval:
@bocharsky-bw The problem with this "Symfony + Docker" thing is that a Symfony docker stack would need at least 2 containers : php-fpm + nginx, and there's some dichotomy between devs about "apache or nginx", but also "one or multiple containers", "do we need mysql or mariadb", etc.
 
I don't think that the @Symfony team is ready to set up a "standard recommendation for Docker" or any other virtualization tool (like homestead with vagrant or just a simple virtualbox container).

This is stuff that also has to be maintained, and as everything grows up very fast in our world, the core team would have to discuss about these things very deeply before making any changes in the Standard Edition to adapt to **every** environment, including virtualization or container-based stacks.

That's why I'm 👎 for this change : the Standard Edition is "not ready yet" for these advanced stacks. Even if lots of devs are using them.

dunglas:
For information, we maintain a basic Docker setup for API Platform (it works for any Symfony project): https://github.com/api-platform/api-platform

Pierstoval:
As Symfony 4 & Flex are coming, I think we can close this: new development practices will help in maintaining a custom development environment or deployment process.

PR #1032, opened 2016-12-14
Update sample sqlite path according to symfony 3 directory structure

chalasr:
I think sqlite databases should be stored in `var/` as it is storage file


xabbuh:
👍

chalasr:
The change has been merged in the docs.
ping @fabpot

fabpot:
Thank you @chalasr.

PR #1033, opened 2016-12-14
Fix ability to uncomment the sqlite path easily

chalasr:
So both key and value are aligned to the previous config line using the uncomment shortcut of IDEs


fabpot:
Thank you @chalasr.

PR #1034, opened 2016-12-16
[do not merge] Change the default directory structure and use new features of 3.3

dunglas:
This PR leverages is part of the 0 config initiative. It leverages the numerous improvements done in 3.3 and [ActionBundle](https://github.com/dunglas/DunglasActionBundle):

* The app is now bundle less: all files are now directly in the `src`directory, no `AppBundle` anymore
* Any object can now be injected as a service in controllers (using constructor, getter or setter injection) **without** having to define it as a service in `services.yml` thanks to autowiring
* Common controller's proxy methods are injected using a trait (`ControllerTrait`)
* Getter injection is used in the controller (lazyness everywhere)


app/config/services.yml
```diff
@@ -4,6 +4,5 @@ parameters:
 #    parameter_name: value
 
 services:
-#    service_name:
-#        class: AppBundle\Directory\ClassName
+#    'My\ClassName':
```
mickaelandrieu: One question: in your tests, how can you replace this service with another one (to use stubs for instance)?

stof: like any other service: by defining the class name explicitly:

```yaml
services:
    'My\ClassName':
        class: 'My\TestClassName'
```

mickaelandrieu: Yeah, this is what I'm afraid of: you're forced to mix both syntaxes :/

stof: what both syntaxes ? We don't have 2 syntaxes. We just made ``class`` default to the id when it is omitted. The id of your service is just ``My\ClassName`` in this case.
You are free to keep using other ids if you prefer btw. Services will not be forced to use a class name as id.

mickaelandrieu: > Services will not be forced to use a class name as id.
Ok, I'm very afraid to be forced to use it, please keep it possible even for Sf4 please :)

xabbuh: Forcing to use the class name does not make sense as you would then not be able to register more than one service for a class.

fabpot: Keep in mind that this new "feature" si optional, don't use it if you don't like it. Anyway, IMHO, if you need to change the class of a service for testing, you are doing something wrong.

mickaelandrieu: Thanks @fabpot. 
Agree with you at some point, it could be useful to set new services in "test" environement when you want a "null" strategy like using in carsonbot for instance (https://github.com/carsonbot/carsonbot/blob/master/app/config/services_test.yml#L7). Not the topic, but if it's wrong I'd like to know why :)

app/config/routing.yml
```diff
@@ -1,3 +1,3 @@
-app:
-    resource: "@AppBundle/Controller/"
+action:
+    resource: "../../src/Controller/"
```
fabpot: This works by chance :) see symfony/symfony#21231

dunglas:
Now includes updated versions of:

* symfony/symfony#20973
* symfony/symfony#21031
* symfony/symfony#18193
* ActionBundle

It's ready to test.

lunika:
I really like the idea of Bundleless symfon applications but with this structure how we should register some features like : 

- compilerPass, I usually put them in `AppBundle::build` method
- custom configuration : a `Configuration` class is put in `DependencInjection` directroy and used in the `AppExtension` class where the config is computed and injected in the container (for custom parameters for example)

stof:
> custom configuration : a Configuration class is put in DependencInjection directroy and used in the AppExtension class where the config is computed and injected in the container (for custom parameters for example)

if you need a semantic configuration, you will need a bundle (well, technically, you need a DI extension, but adding one without a bundle becomes painful even though it is possible, as non-bundle extensions are called only when they have some config explicitly)

> compilerPass, I usually put them in AppBundle::build method

you can override ``Kernel::prepareContainer`` to register additional compiler passes (don't forget to start with a ``parent`` call, otherwise you will break any bundle-provided DI features)

chalasr:
I personally agree with all the goals of 3.3 (it took me some time to be ok with some of them though, but here we are). 
However, I do think bundles are a great feature that is imo not something hard to understand when starting to use symfony, even more, it gives an idea of what is separation of concerns and a good opportunity to start applying it for then, when becoming pretty comfortable with symfony, start moving stuff outside of bundles. 
Nobody forces you to use bundles as well as nobody forces you to use autowiring, but you can at least find an example of this great concept and start playing with it as you wish thanks to the `AppBundle`. Manipulating bundles in its own app also makes easy to understand how to manipulate and write 3rd party ones, thus reuse instead of reinvent.
So, do we really want to remove the concept of bundles from the default structure? 

(It's just a question, do not see any offence, I personally love bundle-less apps, but only after after having used bundles for years since these ones given me a structured way to organize my code and, I fairly trust it, helped to make me a better developer)

pocky:
Hello folks,

I love this PR but I also have a question related to my freelance position with lot of legacy-to-(symfony|microkernel|silex) projects. If there is no more AppBundle and the "best practices" encourage us to add our configuration, views, etc in `app`, why you don’t simply move `src` to `app/src` ?

There is (at least) three advantages to do this for DX and/or project :
- I know a lot of people hate this but (even if I don’t like it too), renaming the app folder and add another one for multi-app (ok you will have problems with `ScriptHandler` but multi-app is not standard so...)
- For monolithic projects, add your frontend application (vue.js for example but it works with a lot of things) aside to your backend application
- If an application is growing up : cut/paste/done.

And for @chalasr because his question is a true one, rename the actual `src`to `bundle` could be an answer.

And again, I love this PR :D

dunglas:
The goal of this PR is to experiment with automatic service registration and bundle less apps. However the final target is Symfony Flex (it's why this PR has the `[do not merge]` flag). Let's see what it looks like first.

dunglas:
I've updated this PR to use new features of 3.3 instead of ActionBundle:

* controllers and commands are still automatically discovered and registered as services
* controllers and commands are still autowired, you can typehint any class in its constructor and use it as a service without having to register it manually as a service (to create a new service: create a class somewhere in `src`, typehint it in your constructor or anywhere else, it will be automatically instantiated and injected)
* for newcomers and performance sensitive apps, the `ControllerTrait` can be used to automatically inject common services (same as the current `Controller` class), but is not mandatory (if you want to cherry-pick deps of your controller, just typehint services you need in its constructor)

I've also updated the directory (just a proposal):

* `conf/`: config files
  *  `dev`: conf overriding for the `dev` env
  *  `prod`: conf overriding for the `prod` env
  *  `test`: conf overriding for the `test` env
* `src/`: all PHP code (including the kernel)
  * **no more `AppBundle`**
  * `Controller/`: Controllers
* `templates/`: Twig templates (can be removed if you don't use Twig - e.g. API Platform)

Unchanged:

* `bin/`
* `var/`
* `vendor/`
* `tests/`
* `web/`

ruudk:
Nice! I have a question: Why not do `config/` instead of `conf/`?

dunglas:
@ruudk It's shorter and maybe more common, but I've nothing against `config`.

lyrixx:
@dunglas As you are breaking everything, could we broke also:
`app/Resources/views/base.html.twig → templates/base.html.twig` to `app/Resources/views/base.html.twig → templates/layout.html.twig` (base->layout) ?

mickaelandrieu:
@dunglas really interesting, this also need a MASSIVE update of docs and probably a new "Best practices" standard or all apps will become a mess ^^

Is this intended to replace the actual Symfony full edition? Or is Symfony flex a new "edition"?

dunglas:
@mickaelandrieu it's just a personal experiment for now, nothing official (it's why the PR title contains *[do not merge]*

dunglas:
@lyrixx done

joelwurtz:
Where does your ressources directory will go with this structure ? (translations / assets / ...)

yceruto:
According to the new structure (purposed) the configuration path should be
```
public function registerContainerConfiguration(LoaderInterface $loader)
{
    $loader->load($this->getRootDir().'/../conf/'.$this->getEnvironment().'/config.yml');
}
```
(i.e. `/conf/dev/config.yml`)

src/Controller/Hello.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\Routing\Annotation\Route;
+
+class Hello
+{
+    use ControllerTrait;
+
+    /**
+     * @Route("/", name="homepage")
+     */
+    public function __invoke(Request $request): Response
```
yceruto: Is mandatory to use `__invoke()` method here? if you are using route annotation, why not use a common method name? (such as `index()` for example)

Pierstoval: It's to propose the concept that a controller represents a single action.

Using `index()` would encourage to create _other_ methods, whereas using `__invoke()` will encourage you to create one single route bound to this controller.

Why not recommend (again?) the use of xml/yml for routing instead of using annotations? This would seriously decouple from the SensioFrameworkExtraBundle, and also reinforce the "single-action controller" concept

nicolas-grekas: Why do you consider that an annotation is coupling? You can perfectly ignore it and do otherwise if you'd like to. It's plain descriptive only to me. "Use if you want". Thus no coupling at all.

maximecolin:
I already use quite the same architecture in my projects (as far as the current Symfony release allow me to go). I will be pleased to see Symfony go this way.

src/Controller/Hello.php
yceruto: > Using index() would encourage to create other methods, whereas using __invoke() will encourage you to create one single route bound to this controller.

It is confusing to me, `Hello` is a controller (`src/Controller/Hello.php`), a controller has action**s** (in most cases), if the concept is "A controller represents a single action" it mean that I need a controller per action? so the path shouldn't be `src/Action/Hello.php` then? We need create a php file per action then? :/

jvasseur: For Symfony, a controller is a callback so in fact event in the traditional Symfony structure the controllers are the methods not the classes, they are just badly named.

mickaelandrieu: I'm not sure we should encourage the "single-action-controller" principle, from a DX point of view this is terrible!

Pierstoval: @nicolas-grekas You're totally right, I'm more talking about _"Preventing encouragement of coupling"_ rather than _real_ coupling.

@yceruto You're right, and as @jvasseur says, in MVC, a controller is supposed to get the user action, maybe use model and view layers, and return a result of the user actions. In Symfony, a controller is a class that can do lots of actions. It's a bit counter-intuitive.

@mickaelandrieu This is the purpose of the ADR (Action-Domain-Responder) pattern (a MVC refinement attempt) which has been proposed by @dunglas for a long time now for Symfony (for example with his [DunglasActionBundle](https://github.com/dunglas/DunglasActionBundle))
I find this pattern better than MVC because it's less fuzzy about how to implement actions, more flexible for the Domain part and a Responder could be anything, whereas MVC talks about views (but in APIs you don't technically have any View layer, for example).
_In fine_, Symfony won't be refined with only-invokable controllers (this would be a HUGE BC break), but encouraging people to more decouple their business logic is better for reusability and organizing your different classes.

chalasr: A controller is a callable that returns a response (i.e. `_controller` in routing), also called actions. Not a symfony specificity to me.
Having only one controller per class does not impact DX IMHO, even less questions at naming time (find a name covering several responsibilities vs only one).

yceruto: Well, a (controller) class is a (callback) callable "only" if this one implements the `__invoke()` method otherwise it's not.

But what I'm trying to say here is that `@route()` annotation is setting up the `_controller` as `App\Controller\Hello::__invoke` directly, so using `__invoke()` method is useless here (i.e. does not make any magic calls).

Pierstoval: @yceruto Indeed, it does not involve magic call, but as said, it _helps_ to realize that what Symfony calls a _controller_ is actually _any PHP callable_. Somehow this concept is not totally obvious for everyone, and as said above, @dunglas proposes ADR pattern, so using an invokable class is basically the best solution for this, even using the `@Route` annotation.
 
And I'd personally get rid of the annotation in favor of yml routing :trollface:

stof: @yceruto in HttpKernel, the controller is not the class. It is always the callable.

A controller cannot have actions. A single class can contain multiple controllers. Controllers are **always** methods in Symfony (unless you go back to symfony 1 where the term `controller` had a different meaning).
Btw, when using Silex, it is very common to avoid using classes entirely and use closures instead (which cannot be done in the fullstack framework as it does not play well with caching)

mnapoli:
Where would you put resources other than views and config? For example translations, data files, assets, fixture files, etc.

I played with Puli a lot and I have to say I like a lot they're recommended layout: having a `res/` folder containing all resource files. That complements very well the `src/` folder (sources VS resources). Here is an example:

```
bin/
res/
    config/
       dev/
       prod/
       test/
    views/
    translations/
    public/
src/
var/
vendor/
web/
```

Anyway that's just feedback from having played with that layout.

Regarding the multi-app question (@pocky) to me all those folders are related to the application. If you want to create another app, that other app will have its own tools (`bin`), resources (config, views, …), sources, vendors, etc. So moving `src/` into `app/` is not enough, you would have to move all those folders into `app/`, which in the end defeats the point of having the `app/` folder at all since we are already in the application folder (i.e. if you have another app, just put those apps in 2 separate sub-folders).

iltar:
@lunika 
> compilerPass, I usually put them in `AppBundle::build` method

See: symfony/symfony#20107

src/Controller/Hello.php
yceruto: @stof thanks for the explanation, to be clearer, we are defined two possible controllers here (at the same time):

 1. `'App\Controller\Hello'`
 2. `'App\Controller\Hello::__invoke'`

Currently we are using the second controller with the appearance to use the first one, so my suggestion here is to use a better method name if we use already `@route()` annotation, otherwise let's define the controller into some `routing.yml` to use the first one only.

Now in Symfony it is not possible to declare this type of route as annotation but could be good to have something like this:
```
/**
 * @Route("/", name="homepage")
 */
class Hello
{
    public function __invoke()
    {
        return new Response('Hello World!');
    }
}
```

Actually equal to: 

```
homepage:
    path: /
    defaults: { _controller: src\Controller\Hello }
```

maximecolin:
@mnapoli IMHO views and config are not resources. Views are almost sources and config, well it's config. It won't bother me to have a translations repertory at the same level as `src/`.

ogizanagi:
@maximecolin, @mnapoli : IMHO views and configs are considered resources in a PHP project, as soon as it needs to be located and is consumed by your app. 
I actually like the `res` folder convention, and given the changes done in this PR, I'll find natural to find config, views, translations (,...) in it.

mnapoli:
What's also to consider in this change of putting all resources together (and then I'll stop talking about it) is that it's a first step towards universal resource location (which is actually Puli's goal).

Everything in `src/` can be loaded/autoloaded using PSR-4. For the rest, a `res/` folder could maybe enable a standard in resource location? Just throwing that idea in there :)

src/Controller/Hello.php
Pierstoval: @yceruto Your suggestion to put the route annotation on the class for invokable controller class is very interesting, and could be the object of an issue/PR in the SensioFrameworkExtraBundle repo 😄

Pierstoval:
And to add my brick in the wall, a `res/` folder could be shared between different services and loaded via resource locators in different ways, like Puli proposes 😏

pocky:
@Pierstoval Puli is a great tool but we have to be careful. The project is not very active, it is still in beta, it is much more than a Resource Locator and, unless this has changed, Puli is not yet compatible with Twig 2.x.

So two options, create a Resource Locator _à la Puli_ or contribute to the project.

BTW, I ❤️ Puli and this PR too.

jrobeson:
is there any way to get rid of `conf/autoload.php`? doesn't feel like the the right place.   Would be nice if composer could handle the annotation registry registration automatically.

pocky:
@jrobeson Good point.

My current directory structure for a single app :
image

maximecolin:
@ogizanagi Ok, I see what you mean. I'm just quite not confortable with the idea views are not code sources in the same way php files are. But it make more sens with this definition of resources.

iltar:
@dunglas what things like typescript, less, javascript etc, things that need pre-processing? Currently I use `app/Resources/assets`, but this feels wrong.

kbond:
I have been using the following structure lately and quite like it:

```
.
├── bin
├── config
├── etc
├── resources
│   ├── assets
│   │   ├── images
│   │   ├── js
│   │   └── sass
│   ├── fixtures
│   ├── migrations
│   ├── translations
│   └── views
├── src
│   ├── AppCache.php
│   ├── AppKernel.php
│   └── autoload.php
├── var
├── vendor
└── web
```

src/Controller/Hello.php
sstok: Pure MVC was never meant to be used on the web, ADR better describes how a web application works.
And for invocable Http-endpoints (which is technically what these are) I prefer Action instead Controller as it better suites the intention.

sstok:
Whats the etc about? Docker and build-stuff?

kbond:
@sstok yeah, anything that doesn't really fit in the other dirs

src/AppKernel.php
```diff
@@ -0,0 +1,67 @@
...
+                $bundles[] = new WebServerBundle();
+            }
+        }
+
+        return $bundles;
+    }
+
+    public function getRootDir()
+    {
+        return __DIR__;
```
GromNaN: since templates and config are in the parent directory, why not setting the root dir at the root of the repository ?

templates/default/index.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html.twig' %}
+{% extends 'layout.html.twig' %}
```
GromNaN: I do prefer "base" over "layout". The layout refers to the style.

src/Controller/Hello.php
yceruto: Now my proposal is in progress: [[Routing][DX] Add full route definition for invokable controller/class](https://github.com/symfony/symfony/pull/21723)

src/AppKernel.php
Pierstoval: Because it's **kernel**'s root dir, not **application** root dir

templates/default/index.html.twig
weaverryan: Also, let's not change anything unnecessarily.

conf/services.yml
```diff
@@ -0,0 +1,40 @@
...
+
+        Twig_ExtensionInterface:
+            tags: ['twig.extension']
+
+        Symfony\Component\EventDispatcher\EventSubscriberInterface:
+            tags: ['kernel.event_subscriber']
+
+        Symfony\Component\Form\FormTypeInterface:
+            tags: ['form.type']
+            public: true  # Mandatory
```
BPScott: Can this be removed now that symfony/symfony#21690 is merged and thus form types can be private?

xabbuh: 👍

stof:
> is there any way to get rid of conf/autoload.php? doesn't feel like the the right place. Would be nice if composer could handle the annotation registry registration automatically.

@jrobeson see https://github.com/symfony/symfony/pull/21837 which is working on it.

Pierstoval:
Just an idea: can't the "action edition" become a new official Symfony edition instead of replacing the Standard Edition?

Symfony CMF already makes updates to his full-stack application from time to time, all based on standard edition, and they seem to agree with it (or maybe it's just me who thinks this 😛 ).

I mean, it has been a bit discussed that the installer could install different Symfony distributions than Demo and Standard Edition (see symfony/symfony-installer#39 ), but actually, the "action pattern" one could easily be one of the different editions.

And it could engage the development process in the installer & the Symfony team on _"How to define a Symfony edition"_, etc.

Obviously, more editions = more maintenance, but as the "action" one will start with Symfony 3.3, this is less maintenance than the SE 😆

stof:
@Pierstoval having multiple official editions has a huge impact on the documentation. So I would not make it an official extension today, especially given that this PR is clearly experimental currently (and relies on unmerged Symfony PRs).
The big maintenance overhead of adding a new edition is not in the installer (where we could easily add support for another one) or in the distribution channel (the building of the SE archive is automated, so this could probably be reused). It is in the need to maintain the documentation twice due to different ways to structure the project.

Pierstoval:
I partially agree. Actually, I agree on the fact that there is an impact on the documentation.

But actually, a new Symfony edition is maintained by its maintainers, and so should be its documentation, what do you think?

fabpot:
Symfony Distributions are dead. They won't exist in Symfony 4. We won't need them anymore. This will be replaced with Symfony Flex instead. I really have to blog about that as not everybody was at SymfonyCon in Berlin. Will try to do that ASAP to give more background.

Pierstoval:
> Symfony Distributions are dead. They won't exist in Symfony 4.

That's great news! 👍

stof:
> I really have to blog about that as not everybody was at SymfonyCon in Berlin. Will try to do that ASAP to give more background.

would be great, especially as there is nothing about your keynote available AFAIK: no slides, no video (last time I checked).

fabpot:
Closing this one. Flex is on its way now :)

teohhanhui:
So you're confident that Flex will resolve all of these issues, and that everyone will embrace it? :smile:

PR #1036, opened 2016-12-28
Sort Composer packages automatically

javiereguiluz:
I do this in all my projects and Laravel added this too recently (https://github.com/laravel/framework/pull/16843) but I don't know if others will agree 😄

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,8 @@
     "config": {
         "platform": {
             "php": "5.5.9"
-        }
+        },
+        "sort-packages": true
     },
     "extra": {
         "symfony-app-dir": "app",
--- a/composer.json
+++ b/composer.json
@@ -12,16 +12,16 @@
     },
     "require": {
         "php": ">=5.5.9",
-        "symfony/symfony": "3.3.x-dev",
-        "doctrine/orm": "^2.5",
         "doctrine/doctrine-bundle": "^1.6",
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
In this case, I think you have to reorder existing packages as well in this PR.

HeahDude:
👍

dunglas:
👍

bocharsky-bw:
What about to sort registered bundles in `AppKernel.php` in addition to this suggestion?
P.S. I'm not sure this idea worth to be opened in another issue.

lyrixx:
:+1:

fabpot:
How does that work? I've just added this config to my composer.json, but running composer install or composer update does not sort my composer.json deps.

chalasr:
@fabpot indeed, running `install` or `update` doesn't seem to trigger it but `require` does.

fabpot:
@chalasr Ok, sounds a bit counter-intuitive. I would have expected `update` to reorder my deps as well.

chalasr:
I agree, that could deserve a PR on composer.

stof:
@fabpot ``composer update`` is not allowed to alter the composer.json file (and ``composer install`` is not even allowed to alter the lock file)

fabpot:
I've done that as of 2.7 and merged everything up.

PR #1037, opened 2016-12-29
Fix deprecated swiftmailer parameter

piotrantosik:


xabbuh:
IMO you should then also update the version constraint in `composer.json` from `~2.3` to `~2.3,>=2.3.10`.

piotrantosik:
@xabbuh You're right, this commit should be base on 2.3 branch - but 2.3 will be merged into 2.7?

bocharsky-bw:
@piotrantosik He means to change this line: https://github.com/symfony/symfony-standard/blob/master/composer.json#L19

PR #1038, opened 2016-12-29
[Fix] Notice: A non well formed numeric value encountered

anddezr:
realpath_cache_size can be with or without unit. 
Remove last character ($unit) from string $size if not all characters in $size are numerical.


ogizanagi:
This should be fixed directly at the source, in the [SensioDistributionBundle](https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Resources/skeleton/app/SymfonyRequirements.php#L779-L794), otherwise it'll be overwritten on next update.

HeahDude:
Let's close here, since there is now https://github.com/sensiolabs/SensioDistributionBundle/pull/303.

PR #1039, opened 2016-12-30
Fix deprecated swiftmailer parameter

piotrantosik:
fixed pr https://github.com/symfony/symfony-standard/pull/1037 @xabbuh please review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.4",
         "symfony/assetic-bundle": "~2.3",
-        "symfony/swiftmailer-bundle": "~2.3",
+        "symfony/swiftmailer-bundle": "~2.3,>=2.3.10",
         "symfony/monolog-bundle": "~2.11.3",
         "sensio/distribution-bundle": "~4.0",
         "sensio/framework-extra-bundle": "^3.0.2",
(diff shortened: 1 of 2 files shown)


xabbuh (approved):

fabpot:
Thank you @piotrantosik.

composer.json
```diff
@@ -12,7 +12,7 @@
         "doctrine/orm": "^2.4.8",
         "doctrine/doctrine-bundle": "~1.4",
         "symfony/assetic-bundle": "~2.3",
-        "symfony/swiftmailer-bundle": "~2.3",
+        "symfony/swiftmailer-bundle": "~2.3,>=2.3.10",
```
bocharsky-bw: Why not just `~2.3.10`?

xabbuh: `~2.3.10` would forbid to install `2.4.0`, for example, once it got released.

bocharsky-bw: @xabbuh Ah, it makes sense. But `^2.3.10` would allow `2.4.0`, right?

xabbuh: Yes, that would be equivalent.

bocharsky-bw: @xabbuh Thanks and happy New Year! :)

xabbuh: @bocharsky-bw Happy new year for you too! :)

Pierstoval: Yep I also recommend `^2.3.10` which is nicer :)

fabpot: done in more recent versions of SE

PR #1040, opened 2017-01-04
Update symfony/monolog-bundle to 3.x

pierredup:
Currently when using Symfony >= 2.8.14 with monolog-bundle 2.11.3, no logs are shown in the profiler. When using monolog-bundle 2.12, the logs appear again.

I haven't tested on 2.7 yet, but this change should work fine for 2.7 as well.

Related: symfony/symfony#21014

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "doctrine/doctrine-bundle": "~1.4",
         "symfony/assetic-bundle": "~2.3",
         "symfony/swiftmailer-bundle": "~2.3,>=2.3.10",
-        "symfony/monolog-bundle": "~2.12",
+        "symfony/monolog-bundle": "^3.0.2",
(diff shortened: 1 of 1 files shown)


stof:
``~2.11.3`` is indeed a bad constraint as it forbids getting the latest releases.

I would even migrate to 3.0.2+ to have the new features.

pierredup:
The same issue happens when using 3.0.1+ (I haven't tested on 3.0.2 yet, will do so shortly).
So if you want to migrate to 3.x, it needs to be locked on 3.0.0 for the time being

jordisala1991:
It should work, atleast for the changelog.. is reverting the change that was causing the issue

pierredup:
I've tested with 3.0.2 and it works fine, I have updated the PR to use 3.0.2+

fabpot:
Thank you @pierredup.

PR #1042, opened 2017-01-06
Register WebServerBundle

### LINKED ISSUE #1041: Wire WebServerbundle for 3.3
Reminder: 
Once symfony/symfony 3.3  is released, we should register the WebServerBundle (AppKernel) when updating vendors, so the `server:*` commands are available. Same for the [symfony-demo](/symfony/symfony-demo) repository and any "demo" project which makes use of these commands.

chalasr:
Updates vendors which was locked to 3.2.1 and registers the WebServerBundle. 
Closes #1041

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -23,6 +23,10 @@ class AppKernel extends Kernel
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
+
+            if ('dev' === $this->getEnvironment()) {
+                $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
+            }
         }
 
         return $bundles;
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,6 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "3749ec07f0cd5aa8a58d5252a2c19ca4",
     "content-hash": "dbdc24fc713f6a6da88dcdff23201f1b",
     "packages": [
         {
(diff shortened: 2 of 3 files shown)


app/AppKernel.php
```diff
@@ -13,6 +13,7 @@ public function registerBundles()
             new Symfony\Bundle\TwigBundle\TwigBundle(),
             new Symfony\Bundle\MonologBundle\MonologBundle(),
             new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
+            new Symfony\Bundle\WebServerBundle\WebServerBundle(),
```
fabpot: Should probably be registered only for the dev environment.

chalasr: Of course... fixed

app/AppKernel.php
```diff
@@ -21,6 +21,7 @@ public function registerBundles()
         if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
+            $bundles[] = new Symfony\Bundle\WebServerBundle\WebServerBundle();
```
fabpot: I would really add it just for the dev env, it does not make sense to register it for the test one, right?

composer.lock
```diff
@@ -4,7 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "e8806838d7235c3d2ab386293836949f",
+    "hash": "3749ec07f0cd5aa8a58d5252a2c19ca4",
+    "content-hash": "dbdc24fc713f6a6da88dcdff23201f1b",
```
stof: please update your composer version to use composer 1.3, instead of reverting the lock file to things created by an older version (different date format and the old ``hash``), and then regenerate the lock with ``composer update --lock``

chalasr: done

app/AppKernel.php
chalasr: Good point. Wouldn't the same be relevant for SensioGeneratorBundle?

fabpot: Indeed

chalasr: Shall we make the change here?

fabpot:
Thank you @chalasr.

PR #1044, opened 2017-01-26
Don't log console errors in console output

chalasr:
Since https://github.com/symfony/symfony/pull/21003, console errors are logged, and since the `console` channel is not ignored by the monolog console handler, the log is written to stderr.

IMHO it's useless since the ran command already write the message into, it's just having two traces for the same error, which doesn't make sense to me. 

I propose to fix that.

__Before__
before

__After__
after

If one wants the trace, then one can run the command in verbose mode or look at its log file.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -20,7 +20,7 @@ monolog:
             channels: [!event]
         console:
             type:   console
-            channels: [!event, !doctrine]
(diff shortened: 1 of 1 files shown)


chalasr:
cc @jameshalsall

jameshalsall:
Great catch @chalasr :) thanks

fabpot:
Thank you @chalasr.

PR #1045, opened 2017-02-16
explicitly declare supported Twig versions

xabbuh:
Not being explicit which Twig versions the application supports mean
that users will get new major Twig releases when they become stable and
composer update is run. This can lead to undesired effects when the new
ajor release is breaking backwards compatibility.


fabpot:
I think this needs to be done in 2.7 as well.

xabbuh:
Okay, I changed the target branch.

fabpot:
Thank you @xabbuh.

PR #1046, opened 2017-02-22
Fix "!event" deprecation

dunglas:


stof:
I would use single quotes. Single-quoted YAML scalars are simpler, because there are no special chars inside them (except the single quote itself)

dunglas:
@stof done (I've also replaced all existing double quotes by single quotes for the sake of consistency).

fabpot:
Shouldn't this be done in 2.7 instead?

dunglas:
Why not, but the deprecation is triggered only since 3.3.

stof:
Replacing existing quotes should indeed be done in 2.7, to make merging branches easier

dunglas:
* Switched to 2.7
* Uniformized indentation (no align)
* Unifomized comments

fabpot:
Thank you @dunglas.

PR #1047, opened 2017-02-24
updated instruction wording in comment

httpoz:


fabpot:
Good catch, thanks @httpoz.

PR #1048, opened 2017-02-28
Improve PSR-4 autoloading by default

Pierstoval:
TL;DR: (updated Marth 3rd)
Before, I proposed to use any optimization by default in `composer.json` (complementary to PSR-4 autoload for the AppBundle).

I took note of the different arguments and removed loaders optimization because it is not suitable for dev environments, but kept AppBundle PSR-4 explicit load because it encourages PSR-4 best practices of autoload.

--- 

As seen on symfony/symfony-demo#490 (and it's certainly a well-known issue for developers using PHP Inspections EA Extended plugin for PHPStorm), having empty namespaces in `psr-4` rules has an impact on performances.

Using this by default on the Standard Edition may encourage developers to add PSR-4 rules if they create new namespaces (for example if they want to create bundleless apps or other bundles, or components inside the same app, etc.).

What do you think about putting this in the SE?

Also, for best autoloading performances, it is recommended to use `classmap-authoritative` options (which adds `optimize-autoloader` implicitly). I added it in the `composer.json` but I'm not sure whether it's the best option to propose to Symfony newcomers (because `composer install` and `update` takes more time to execute).

For this have a question: should we recommend `classmap-authoritative` or `optimize-autoloader` (or nothing ❔) by default ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,9 @@
     "type": "project",
     "description": "The \"Symfony Standard Edition\" distribution",
     "autoload": {
-        "psr-4": { "": "src/" }
+        "psr-4": {
+            "AppBundle\\": "src/AppBundle"
+        }
     },
     "require": {
         "php": ">=5.3.9",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "bdde043c25314bfbc2727929457166a0",
+    "content-hash": "7179f4c120649326ae8dab4da26c6db9",
     "packages": [
         {
             "name": "doctrine/annotations",


stof:
Optimizing the autoloader should not be in the project-level composer.json. This is not a setting that you want to use in development. ``classmap-authoritative`` is particularly annoying in dev as it requires redumping the autoloader each time you add a class). ``optimize-autoloader`` is a bit better, but it still lead to weird errors when you delete a class (as the composer classloader will still assume the file exist and will try to require it when asking for the class).

Such settings should be turned on only in prod environments (and ``classmap-authoritative`` requires makign sure that your project works fine with it, which may not always be the case depending of what your code does)

stof:
To be clear, ``classmap-authoritative`` should be used in prod as soon as you can. And ``optimize-autoloader`` is just a must-have in prod. Not using it means you don't care about performance at all, as it is a performance boost without drawback (it has drawbacks in dev, but not in prod)

Pierstoval:
So you recommend nor `classmap-authoritative` nor `optimize-autoloader` for the SE?

stof:
@Pierstoval I never recommend putting them in the project itself (unless you never develop in it).
this setting should either be enabled in the system-level config on prod server (in ``$COMPOSER_HOME/config.json``) or through the CLI flags in your deployment scripts.

Pierstoval:
@stof:
> `optimize-autoloader` is a bit better, but it still lead to weird errors when you delete a class (as the composer classloader will still assume the file exist and will try to require it when asking for the class).

I find this argument very interesting: if you delete a class, it shouldn't be used in your application, therefore the `class not found` error would still be triggered by the autoloader, wouldn't it?

In this case, it means that `optimize-autoloader` would trigger errors only for mistakes that are already handled without optimizing the autoloader, so I would still suggest we may use it by default (instead of `classmap-authoritative` that sounds more production-specific)

stof:
@Pierstoval there are case like ``class_exists`` where passing a non-existent class to the autoloader is expected.
An outdated optimized autoloader will not be silent in this case, but will require the removed file.

Pierstoval:
Hmm... That makes me re-think about default autoload optimization in app. I may remove this part in this PR if it sounds too much for the SE.

By the way, what about the `psr-4` changes? 😃

bocharsky-bw:
I suppose `composer.lock` should be updated.

Pierstoval:
I removed the `classmap-authoritative` part and updated the lock file hash

stof:
Composer now has a documentation page about optimizing the autoloader for production: https://getcomposer.org/doc/articles/autoloader-optimization.md

Pierstoval:
@stof Great! Then, I guess the Symfony documentation about deployment best practices can also link this article from Composer's website, and we can still review explicit PSR-4 autoload rule for AppBundle, I tend to continue thinking it's a good practice 😃

stof:
@Pierstoval linking to the composer doc from the Symfony one is the plan, yes. But we wanted to have the source of truth in the composer doc itself (as it is not specific to Symfony)

Pierstoval:
Then this PR needs to be reviewed 😸

fabpot:
Thank you @Pierstoval.

PR #1052, opened 2017-03-01
added support for the new @required annotation

fabpot:
Needed for symfony/symfony#21763

### DIFF
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -1,5 +1,6 @@
 <?php
 
+use Doctrine\Common\Annotations\AnnotationReader;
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
In fact, "Required" is already listed, so maybe we can recommend that while we suggest doctrine to accept the "required" variant?

fabpot:
I think using `@Required` is a bad idea for the reasons I explained in the original PR.

fabpot:
Closing in favor of #21837 (whatever the solution there, we will get rid of the autoload file here)

PR #1054, opened 2017-03-07
Do not process psr_3 log messages, as it's now done by symfony.

lyrixx:
Related to https://github.com/symfony/symfony/pull/21705

Note: It's not an issue to let monolog doing it, but if monolog do it
symfony can not do it. The best experience is when symfony do it.

So let's disable it.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -20,6 +20,7 @@ monolog:
             channels: ['!event']
         console:
             type:   console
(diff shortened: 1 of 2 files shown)


stof:
this is about getting colors for replaced placeholders, right ?

lyrixx:
@stof Yes. it's all about colors ;)

fabpot:
Thank you @lyrixx.

PR #1055, opened 2017-03-15
Fix sqlite db path in parameters.yml.dist

chalasr:
This should have been done in #1032. ~Adding the path to `.gitignore` was done on symfony-demo, I think we can do it here too.~

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -8,7 +8,7 @@ parameters:
     database_user: root
     database_password: ~
     # You should uncomment this if you want to use pdo_sqlite
-    #database_path: "%kernel.root_dir%/data.db3"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,6 +5,8 @@
 !/var/cache
 /var/cache/*
 !var/cache/.gitkeep
+!/var/data
+/var/data/*
```
Tobion: var/data is already ignored by var/. What's the point of this?

xabbuh: I agree, this only makes sense if we also commit a `.gitkeep` file inside `var/data/`.

chalasr: indeed, reverted

fabpot:
Merging into 3.2 instead.

chalasr:
Rebased on 3.2.

PR #1057, opened 2017-03-24
[Log] Wire server_log handler

lyrixx:
This is the last part of the new `server:log` command.

Maybe a release of the monolog bundle could be better to avoid to require a dev package.
ping @Seldaek 

WDYT?

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -22,6 +22,12 @@ monolog:
             type:   console
             process_psr_3_messages: false
             channels: ['!event', '!doctrine', '!console']
+        # To follow logs in real time, execute the following command:
+        # `bin/console server:log -vv`
+        server_log:
+            type:   server_log
+            process_psr_3_messages: false
+            host: 0:9911
         # uncomment to get logging in your browser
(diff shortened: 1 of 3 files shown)


Seldaek:
monologbundle 3.1.0 is out

app/config/config_dev.yml
```diff
@@ -22,6 +22,12 @@ monolog:
             type:   console
             process_psr_3_messages: false
             channels: ['!event', '!doctrine', '!console']
+        # To follow logs in real time, execute the following command:
+        # `bin/console server:log -vv`
```
kbond: Can we remove the need for `-vv`?

lyrixx: Actually, this is not mandatory ... it's just to get more logs.

kbond: Ah, I see, you only see warnings if excluded.

lyrixx:
I rebased the PR. Now it required the new symfony/monolog-bundle package.

fabpot:
Thank you @lyrixx.

PR #1058, opened 2017-03-24
Replace static calls with actual commands in composer.json

vudaltsov:
Composer now supports using `@php` to run commands. I suggest using this feature in favor of some `ScriptHandler::{method}` calls. This will make changing commands' options and arguments easier.

Also split the `cache:clear` call.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,8 @@
         "symfony-scripts": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
+            "@php bin/console cache:clear --ansi",
+            "@php bin/console assets:install web --symlink --relative --ansi",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget"
         ],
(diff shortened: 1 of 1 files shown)


vudaltsov:
`--ansi` is added to force colorful output

composer.json
```diff
@@ -32,8 +32,9 @@
         "symfony-scripts": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
+            "@php bin/console cache:clear --no-warmup --ansi",
+            "@php bin/console cache:warmup --ansi",
+            "@php bin/console assets:install web --symlink --relative --ansi",
```
HeahDude: I'm 👎 on this change for now as it looses the ability to support extra config from the `composer.json`.

stof:
forcing colors is not good either, as the shell running the command might not support colors

Pierstoval:
I'm also 👎 because composer scripts are using the `extra` composer property (like @HeahDude said), and same for `--ansi`: the projects I deploy are executing these scripts in a shell that do not support colors, so this would result in showing ansi color characters, output becoming less readable. Let the scripts determine automatically whether to show colors or not, like it's already doing now.

composer.json
vudaltsov: My point was to stop using extra config and replace them with real cli arguments and options.

Do you mean that this will be a kind of a BC break?

vudaltsov:
Agree on the colors issue.

stof:
I'm -1 on changing this for Symfony 3.x, as we are currently working on Flex for the next-gen tooling. Let's keep the existing tooling as is (which has less drawbacks than your proposal)

vudaltsov:
Ok, agree with you! Thanx for explanations :)

PR #1063, opened 2017-04-21
remove AsseticBundle after broken merge

### LINKED ISSUE #1062: Why assetic still exists in composer.json?
If I am not mistaken the ascetic bundle was removed from the standard edition, right?

So I am confused when I see it in the composer file.

https://github.com/symfony/symfony-standard/blob/2.8/composer.json#L20

xabbuh:
This will fix #1062.


bocharsky-bw:
I think you have to regenerate `composer.lock` file. Actually, `composer remove symfony/assetic-bundle` should do the trick.

fabpot:
Thank you @xabbuh.

PR #1065, opened 2017-05-02
Use single quotes and project_dir parameter

yceruto:


yceruto:
I just updated other places too. (fabbot fails no related)

fabpot:
Thank you @yceruto.

PR #1066, opened 2017-05-02
remove deprecated %kernel.root_dir% from config.yml

JarJak:
From UPGRADE-3.3.md:
"Deprecated the kernel.root_dir parameter. Use the new kernel.project_dir parameter instead."

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -13,7 +13,7 @@ framework:
     #translator: { fallbacks: ['%locale%'] }
     secret: '%secret%'
     router:
-        resource: '%kernel.root_dir%/config/routing.yml'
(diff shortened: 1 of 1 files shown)


app/config/config.yml
```diff
@@ -13,7 +13,7 @@ framework:
     #translator: { fallbacks: ['%locale%'] }
     secret: '%secret%'
     router:
-        resource: '%kernel.root_dir%/config/routing.yml'
+        resource: '%kernel.project_dir%/../config/routing.yml'
```
xabbuh: should be `resource: '%kernel.project_dir%/app/config/routing.yml'`

JarJak: done

fabpot:
Thank you @JarJak.

PR #1070, opened 2017-05-09
Making *all* classes in src/AppBundle available as services

weaverryan:
Hi guys!

tl;dr; This makes *all* services from `src/AppBundle/` available as services by default.

> This PR depends on symfony/symfony#22680 and symfony/symfony#22665

If this seems crazy to you, it's actually not: it's almost exactly how the system already works :).
By registering *all* of `src/AppBundle`, we will almost undoubtedly register some classes that should *not* be services. But, because the services are private, unless you actually reference them somewhere, they're simply removed. In other words, **we're not *really* importing all services from `src/AppBundle`, we're making all classes in `src/AppBundle` *available* to be used as services**.

Today, thanks to autowiring, if you type-hint a class that is *not* registered as a service, autowiring registers it for you. That means that **the total number of services in the compiled container today versus after this change will be the same**.

* **Today** if you don't explicitly register a service as public and don't type-hint/reference it anywhere, it's never added.
* **After**&nbsp;&nbsp;   if you don't explicitly register a service as public and don't type-hint/reference it somewhere, it's removed from the final container

So, the end result is the same as now. And there are a number of advantages:

**PROS**
* A) **Better developer experience: when a dev creates a new directory**, they don't need to remember to go into `services.yml` and add it. The original directories we added to the SE were a random "guess"... which should feel weird. I think it's because that approach is flawed in general.
* B) **Less WTF and better DX with `autoconfigure`**: now, if I create a new class that extends `Twig_Extension` but forget to import my `Twig` directory, then my extension won't work. After this, we'll find it for sure.
* C) **Clearer for the documentation**: before this change, in many places, we need to say "go to services.yml and add this new directory to your import... unless you already have it... and don't make your line look exactly like our's - you're probably also importing other directories"
* D) **We could deprecating the "autowire auto-registration" code in 3.4** (i.e. when autowiring automatically registers a private services if no instance exists in the container). That means less magic: this is actually *more* explicit. This could be awesome - we could remove some "hacky" code (e.g. symfony/symfony#22306). And, if you type-hinted an `Entity` (for example) in autowiring, instead of the container creating a service for you, it can give you a clear error

In theory, the CON is that this will slow down the container rebuilding as more services are added to `ContainerBuilder` (before being removed later). On one project, I compared the build time without importing everything (138 services added from `src/`) versus importing everything (200 services dded from `src/`). The build time difference was negligible - maybe 10ms difference (both were around 950ms).

Btw, the `exclude` key added in symfony/symfony#22680 is almost not even needed, since unused services are simply removed. But, currently, the `Entity` directory is an exception due to the new "argument resolver" trying to see if it can wire those as services. That could be fixed in the future. But `exclude` is kind of nice, if you want to explicitly safe-guard someone from accidentally type-hinting something from that directory. Again, that's *better* than now... where we will *always* auto-register an Entity if you type-hint it.

Cheers!

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -8,18 +8,28 @@ services:
     _defaults:
         # automatically injects dependencies in your services
         autowire: true
-        # automatically registers your services as commands, form types, etc.
+        # automatically registers your services as commands, event subscribers, etc.
         autoconfigure: true
         # this means you cannot fetch services directly from the container via $container->get()
         # if you need to do this, you can override this setting on individual services
         public: false
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes classes in src/AppBundle available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     AppBundle\:
-        resource: '../../src/AppBundle/{Command,Form,EventSubscriber,Twig,Security}'
+        resource: '../../src/AppBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../src/AppBundle/{Entity,Repository}'
 
+    # controllers are imported separately to make sure they're public
+    # and have a tag that allows actions to type-hint services
     AppBundle\Controller\:
         resource: '../../src/AppBundle/Controller'
         public: true
         tags: ['controller.service_arguments']
+
+    # add more services, or override services that need manual wiring
+    # AppBundle\Service\ExampleService:
+    #     arguments:
+    #         $someArgument: 'some_value'


app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
         # if you need to do this, you can override this setting on individual services
         public: false
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes all classes in src/AppBundle available as services
     # this creates a service per class whose id is the fully-qualified class name
     AppBundle\:
-        resource: '../../src/AppBundle/{Command,Form,EventSubscriber,Twig,Security}'
+        resource: '../../src/AppBundle/*'
```
GuilhemN: This syntax doesn't register sub folders, isn't it ?

app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
...
+        # you can include directories or files
+        # but if a service is unused, it's removed anyways
+        exclude: '^(AppBundle\.php|Entity|Repository)'
 
+    # controllers are imported separately to make sure they're public
+    # and have a tag that makes services
```
GuilhemN: This sentence looks incomplete

app/config/services.yml
weaverryan: It definitely does - the `FileLoader` always passes a `$recursive` flag set to `true` to the glob functionality. But, I totally understand what you're referring to with the normal `**` syntax :)

GuilhemN: Ok, thanks for your explanation ☺

app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
...
+        exclude: '^(AppBundle\.php|Entity|Repository)'
 
+    # controllers are imported separately to make sure they're public
+    # and have a tag that makes allows actions to type-hint services
     AppBundle\Controller\:
         resource: '../../src/AppBundle/Controller'
         public: true
         tags: ['controller.service_arguments']
+
+    # add more service, or override services that need manual wiring
```
GuilhemN: `serviceS`

app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
         # if you need to do this, you can override this setting on individual services
         public: false
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes all classes in src/AppBundle available as services
     # this creates a service per class whose id is the fully-qualified class name
     AppBundle\:
-        resource: '../../src/AppBundle/{Command,Form,EventSubscriber,Twig,Security}'
+        resource: '../../src/AppBundle/*'
+        # you can include directories or files
```
GuilhemN: Should be `exclude`, right ?

app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
...
+        # but if a service is unused, it's removed anyways
+        exclude: '^(AppBundle\.php|Entity|Repository)'
 
+    # controllers are imported separately to make sure they're public
+    # and have a tag that makes allows actions to type-hint services
```
GuilhemN: `makes` should be removed

app/config/services.yml
weaverryan: Ha, thank you for the review! I managed to make a lot of typos... in a very small area :).

GuilhemN: That's why we do reviews :) Thanks for your pr!

app/config/services.yml
```diff
@@ -14,12 +14,22 @@ services:
         # if you need to do this, you can override this setting on individual services
         public: false
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes all classes in src/AppBundle available as services
```
fabpot: I would mention that not *all* classes end up as services

weaverryan:
Awesome! I tweaked the description of importing from `src/` and also updated to the new glob exclude pattern in symfony/symfony#22680

I also found an even more necessary reason for this change. Suppose you have:

```php
class IceCreamGenerator
{
    public function __construct(LoggerInterface $logger) {}
}
```

But you do *not* register this as a service. Instead, you just type-hint this in a controller:
```php
public function indexAction(IceCreamGenerator $generator)
```

This will work: autowiring will auto-register `IceCreamGenerator` as a service. But, suppose now that you add a un-autowireable arg to it:

```php
class IceCreamGenerator
{
    public function __construct(LoggerInterface $logger, $foo) {}
}
```

This is the error the user sees (only when they visit this page) is:

> Controller "AppBundle\Controller\BlogController::indexAction()" requires that you provide a value for the "$generator" argument. Either the argument is nullable and no null value has been provided, no default value has been provided or because there is a non optional argument after this one.

The error is *very* unclear. The reason is due to how the "controller args" feature works: if it has any issue autowiring an arg, it just skips it (it needs to do this, due to things like entities). That's fine, but the error is a problem.

But instead, if you load all services from `src/AppBundle` as I suggest in this PR, then you get correct error (and at compile time):

> Cannot autowire service "AppBundle\Delicious\IceCreamGenerator": argument "$foo" of method "__construct()" must have a type-hint or be given a value explicitly.

app/config/services.yml
```diff
@@ -8,18 +8,28 @@ services:
...
         public: false
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes classes in src/AppBundle available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     AppBundle\:
-        resource: '../../src/AppBundle/{Command,Form,EventSubscriber,Twig,Security}'
+        resource: '../../src/AppBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyways
```
GuilhemN: isn't `anyway` more usual?

fabpot:
Thank you @weaverryan.

fabpot:
see symfony/recipes#65 for the equivalent on Flex.

app/config/services.yml
```diff
@@ -8,18 +8,28 @@ services:
...
 
-    # loads services from whatever directories you want (you can add directories!)
+    # makes classes in src/AppBundle available to be used as services
     # this creates a service per class whose id is the fully-qualified class name
     AppBundle\:
-        resource: '../../src/AppBundle/{Command,Form,EventSubscriber,Twig,Security}'
+        resource: '../../src/AppBundle/*'
+        # you can exclude directories or files
+        # but if a service is unused, it's removed anyway
+        exclude: '../../src/AppBundle/{Entity,Repository}'
```
xabbuh: What about excluding controllers here explicitly too? They are registered below anyway (and excluding them here may prevent some confusion).

weaverryan: I'm mixed. On one hand, I want `exclude` to be a detail that people don't think about or even notice. In fact, the only reason it's even needed is for `Entity`, to avoid a weird situation where the new "controller service args" functionality keeps a reference to them so they're not removed. Adding `Controller` to it makes this setting seem even more important. 

On the other hand, especially in the future if we remove the "auto-register services" functionality from autowiring, exclude will be a nice way to explicitly prevent things from being used as services.

So, I *kind* of think we should not add it. In theory, we could also add `AppBundle.php`... but I think we should keep it minimal instead.

PR #1071, opened 2017-05-10
[minor] Remove spaces in configuration

mayeco:
Remove spaces in configuration


fabpot:
Good catch, thanks @mayeco.

PR #1072, opened 2017-05-10
Use PHP_SAPI instead of php_sapi_name()

javiereguiluz:
In Symfony code we only use PHP_SAPI and https://secure.php.net/manual/en/function.php-sapi-name.php says:

> Note:
> The PHP constant PHP_SAPI has the same value as php_sapi_name().


bocharsky-bw:
👍

fabpot:
Thank you @javiereguiluz.

PR #1074, opened 2017-05-16
app.php and console: no need to declare $loader variable

MacDada:
The var isn't being used anywhere in the file(s).

Because of how "require" works,
$loader is still a global variable.

### DIFF
--- a/app/console
+++ b/app/console
@@ -12,8 +12,7 @@ use Symfony\Component\Debug\Debug;
 
 set_time_limit(0);
 
-/** @var \Composer\Autoload\ClassLoader $loader */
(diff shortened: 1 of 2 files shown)


Tobion:
👍  Should be merged in 2.7 since #964 is also in 2.7 that removed the $loader usage.

MacDada:
@Tobion I'm gonna split it into two PRs, as `2.7` does not have `$loader` in `console`: https://github.com/symfony/symfony-standard/blob/2.7/app/console

fabpot:
Good catch, thanks @MacDada.

Tobion:
@fabpot you forgot app_dev.php when rebasing: #1090

PR #1077, opened 2017-05-18
Use localhost by default for the server log socket.

ChadSikorra:
This fixes the server log on Windows, which is currently causing the dev environment to hang and become unusable:

https://github.com/symfony/symfony/issues/22712

The hostname "0" apparently doesn't work on Windows, but it does on Linux. Though I've honestly never seen "0" used in a connection string before. I figure localhost is the best default route to go. There is still the issue of the ServerLogHandler in the monolog bridge bundle where the underlying issue surfaced.

Basically since "0:9911" was not valid on Windows, the socket was never opened. Yet every time it would attempt to log to the socket again it would go through the same timeout period. Multiply that by all the times that gets called and you'll be waiting a while to get a response... 

Even though this change makes the dev environment usable on Windows, the Server Log handler still has a huge performance impact there. It takes almost 6-8x longer for a page to load in the dev environment now.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -27,7 +27,7 @@ monolog:
         server_log:
             type:   server_log
             process_psr_3_messages: false
-            host: 0:9911
+            host: localhost:9911
         # uncomment to get logging in your browser
         # you may have to allow bigger header sizes in your Web server configuration
         #firephp:


app/config/config_dev.yml
```diff
@@ -27,7 +27,7 @@ monolog:
         server_log:
             type:   server_log
             process_psr_3_messages: false
-            host: 0:9911
+            host: localhost:9911
```
nicolas-grekas, line 30: Does 0.0.0.0 work on Windows?

ChadSikorra: Yes. I guess I wasn't sure what the implication of changing it to localhost would be. I'm guessing 0 gets translated to 0.0.0.0 on 'nix.

ChadSikorra: I take that back. While it helps the dev environment not hang completely, windows will not write to 0.0.0.0 (rendering the `server:log` command useless). On Linux does that just mean write to all interfaces/IPs on the specified TCP port in this context?

ChadSikorra:
Also, whatever is decided here, the `server:log` command will need to be adjusted as well. Currently running that on Windows just throws an exception and the following error message:

> Warning: stream_socket_server(): php_network_getaddresses: getaddrinfo failed: No such host is known.

Running it with `--host localhost:9911` or `--host 127.0.0.1:9911` works. So would `0.0.0.0` in that context I guess.

ChadSikorra:
I found an interesting note in the Windows Socket API error codes that even explicitly states that address 0 is not valid for an address to connect/send to:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668.aspx#WSAEADDRNOTAVAIL

Though if you ping 0 from the command line it returns 0.0.0.0. I guess since `INADDR_ANY` ultimately equals 0 this is some sort of special reserved meaning. Though I cannot find it documented anywhere. But I still think it's an oversight/bug for Linux/Unix networking to even allow a connection to 0.0.0.0 in general, or at the very least something that shouldn't be done.

fabpot:
I think this PR has been automatically closed because I've just removed the master branch. Can you reopen it on 3.3 instead? Thanks.

PR #1080, opened 2017-05-22
Replace localhost with 127.0.0.1

bocharsky-bw:
Using IP address is more robust and we also refer to 127.0.0.1 in [parameters.yml.dist](https://github.com/symfony/symfony-standard/blob/3.3/app/config/parameters.yml.dist) - it will be more consistent with IP address for `server_log.host`.


fabpot:
Thank you @bocharsky-bw.

PR #1082, opened 2017-05-30
DefaultController extends the new AbstractController

vudaltsov:


fabpot:
Thank you @vudaltsov.

fabpot:
Reverted as this has not been tested. Code is broken (`getParameter()` is not available in AbstractController).

vudaltsov:
Yes, my fault to have not taken this into account.

dunglas:
This patch is present in the last tagged version, it means that when you run `symfony new my-project` you get a red error instead of the default page. Can you tag a new version?

PR #1083, opened 2017-06-01
Use HTTPS for links to the docs

bocharsky-bw:
It was done in Symfony Demo https://github.com/symfony/symfony-demo/pull/550

### DIFF
--- a/app/Resources/views/default/index.html.twig
+++ b/app/Resources/views/default/index.html.twig
@@ -36,7 +36,7 @@
                     </svg>
 
                     Read the documentation to learn
-                    <a href="http://symfony.com/doc/{{ constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION')[:3] }}/page_creation.html">
+                    <a href="https://symfony.com/doc/{{ constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION')[:3] }}/page_creation.html">
                         How to create your first page in Symfony
                     </a>
                 </p>
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -4,7 +4,7 @@ imports:
     - { resource: services.yml }
 
 # Put parameters here that don't need to change on each machine where the app is deployed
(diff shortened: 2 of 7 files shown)


app/Resources/views/default/index.html.twig
```diff
@@ -19,7 +19,7 @@
             <div id="next">
                 <h2>What's next?</h2>
                 <p>
-                    <svg id="icon-book" version="1.1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" viewBox="-12.5 9 64 64" enable-background="new -12.5 9 64 64" xml:space="preserve">
+                    <svg id="icon-book" version="1.1" xmlns="https://www.w3.org/2000/svg" x="0px" y="0px" viewBox="-12.5 9 64 64" enable-background="new -12.5 9 64 64" xml:space="preserve">
```
stof: that's wrong. The official XML namespace for SVG does not have `https` in it (the website uses HTTPS, but this does not impact the XML namespace)

phpunit.xml.dist
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.8/phpunit.xsd"
+<phpunit xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
```
stof: that's wrong as well

app/Resources/views/default/index.html.twig
bocharsky-bw: Ah, I followed the link and was redirected to the `https`, that's why I decided to change it.

Thanks, I'll revert it!

fabpot:
Thank you @bocharsky-bw.

PR #1087, opened 2017-06-02
Register WebServerBundle in prod env

maidmaid:
This also allows to run project in prod env, that is usefull to check prod config for example.

```
php bin/console -e prod server:*
```

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -13,6 +13,7 @@ class AppKernel extends Kernel
             new Symfony\Bundle\TwigBundle\TwigBundle(),
             new Symfony\Bundle\MonologBundle\MonologBundle(),
             new Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle(),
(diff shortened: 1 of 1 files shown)


bocharsky-bw:
How do you check prod config with PHP built-in web server?

ghost:
Why would you run this bundle in a production environment? I'd against this pull request.

maidmaid:
In development phase, I sometimes run locally ``php bin/console server:start -e prod`` to check the production things like doctrine cache, monolog handlers or payment configuration. Before 3.3, it was possible to do it *by default*.

chalasr:
I think that's a good thing it is not anymore, the webserver is not meant to ensure that your application runs fine. If you use it for that you can enable it on your side, but that should not be a generalised practice IMO

fabpot:
Keep in mind that this is just the default. Feel free to change it, but default should be about the best practice.

PR #1090, opened 2017-06-16
Remove unused $loader variable

Tobion:
#1074 for app_dev.php which has been forgotten when rebasing


fabpot:
Thank you @Tobion.

PR #1094, opened 2017-06-25
ignore node_modules folder

vaibhavpandeyvpz:
Because now we have @symfony/webpack-encore for asset bundling, I believe the node_modules should be ignored by default.


vaibhavpandeyvpz:
Closing this for a newer pull-request with complete [encore](http://symfony.com/doc/current/frontend/encore/simple-example.html) integration.

PR #1096, opened 2017-06-25
integrate @symfony/webpack-encore into the app

vaibhavpandeyvpz:
These changes integrate [@symfony/webpack-encore](https://github.com/symfony/webpack-encore) as per the docs at [https://symfony.com/doc/current/frontend/encore/simple-example.html](https://symfony.com/doc/current/frontend/encore/simple-example.html).

Did not touch composer's post-install/update hooks to keep it completely optional for those who don't want to make use of it.


fabpot:
I'm not sure we should add it in the Standard Edition. Not everybody would like to use Webpack and Encore. Adding it to as a recipe for Symfony Flex makes more sense IMHO. ping @symfony/deciders

jakzal:
I agree with Fabien. Most of my recent projects don't have frontend at all. A Symfony Flex recipe will make more sense and there's an issues with a PR for that already: https://github.com/symfony/webpack-encore/issues/77

Thank you for your suggestion @vaibhavpandeyvpz nevertheless. I hope you'll continue with the good work.

PR #1115, opened 2017-11-12
Added DataFixtures in Exclude Folders

hectorprats:
I added this folder because if you have DataFixtures in AppKernel in environment Dev mode, it returns error in the deploy when I apply --no-dev config. And we don't need autoload this folder.
So.. is a bug.


Tobion:
👎 Doctrine FixtureBundle is not part of SE. So it does not make sense to add configuration for this.

hectorprats:
I agree, I was wrong. Thank you

PR #1124, opened 2017-12-14
Added a note about Symfony 4 and Symfony Skeleton

### LINKED ISSUE #1121: Mark the repository as deprecated
Could you add a line to readme saying that `symfony-standard` is no longer maintained for `4.0`? Lots of my friends use this repository as a source of truth.

bocharsky-bw:
Resolve #1121

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 Symfony Standard Edition
 ========================
 
+**WARNING**: This distribution does not support Symfony 4. Use [Symfony Skeleton][15]
+instead, which is a minimal Symfony project recommended to create new applications.
+
 Welcome to the Symfony Standard Edition - a fully-functional Symfony
 application that you can use as the skeleton for your new applications.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,3 +73,4 @@ Enjoy!
 [11]: https://symfony.com/doc/3.4/logging.html
 [13]: https://symfony.com/doc/current/bundles/SensioGeneratorBundle/index.html
 [14]: https://symfony.com/doc/current/setup/built_in_web_server.html
+[15]: https://github.com/symfony/skeleton
```
bocharsky-bw: Probably should be pointed to https://github.com/symfony/website-skeleton instead? Looks like it will be a replacement for Symfony SE

weaverryan: Hmm. Or maybe we should link to the Setup page on the docs (and there we can explain the skeleton options). Because if they’re on this repo... they made some mistake and need to start fresh with new info :)

bocharsky-bw: Setup page will give more context 👍

weaverryan:
Ah, now maybe we should just link to the setup page but not mention the “minimalist” skeleton. This is becaue there are now multiple skeletons, so let’s be generic here.

bocharsky-bw:
Done! How about this notice?

Probably "Installing & Setting up the Symfony Framework" page also should be updated to mention "website-skeleton" and others if any

fabpot:
Thank you @bocharsky-bw.

PR #1130, opened 2018-03-16
Fix default timezone test

symple-developer:
symfony/symfony issue: [#26550](https://github.com/symfony/symfony/issues/26550)

Both my php.ini files have correct values for date.timezone var:

```
> php -i | grep "timezone"
Default timezone => America/Sao_Paulo
date.timezone => America/Sao_Paulo => America/Sao_Paulo
```

But Symfony 3.4 requirements test returning error:

```
> php test/bin/symfony_requirements
Configured default timezone "America/Sao_Paulo" must be supported
   by your installation of PHP
   > Your default timezone is not supported by PHP. Check for typos in
   > your php.ini file and have a look at the list of deprecated
   > timezones at http://php.net/manual/en/timezones.others.php.
```

But America/Sao_Paulo isn't deprecated.

### DIFF
--- a/var/SymfonyRequirements.php
+++ b/var/SymfonyRequirements.php
@@ -448,15 +448,10 @@ class SymfonyRequirements extends RequirementCollection
         }
 
         if (false !== $requiredPhpVersion && version_compare($installedPhpVersion, $requiredPhpVersion, '>=')) {
-            $timezones = array();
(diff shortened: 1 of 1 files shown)


Pierstoval:
This should be fixed in https://github.com/sensiolabs/SensioDistributionBundle instead 😄

Tobion:
And https://github.com/symfony/requirements-checker for symfony 4
Please open a PR there

PR #1141, opened 2019-07-08
Switched array() to []

Arman-Hosseini:


lyrixx:
Hello, thanks for your contribution, but this script should be run on the lowest PHP version possible.

Arman-Hosseini:
Hello @lyrixx, and thanks for review this.
Actually **Symfony 3.4 Standard Edition** requires PHP >= **5.5.9** and arrays with using definition [], works with PHP >= **5.4**

bocharsky-bw:
Yes, but the point is to be able to check requirements on PHP version that might be even lower than 5.5.9, then you would get a message that your PHP version does not meet the required one. If we apply tour change - you will see just a fatal error.

fabpot:
@bocharsky-bw is right.

